PR #62, opened 2023-01-18
Add support for Ansys Lab

jgsdavies:


src/ansys/motorcad/core/rpc_client_core.py
```diff
@@ -207,19 +211,62 @@ def __init__(
...
+    def _launch_motorcad_remote(self):
+        """Launch Motor-CAD in Ansys Lab."""
+        pim = pypim.connect()
+
+        self.pim_instance = pim.create_instance(product_name="motorcad")
+        self.pim_instance.wait_for_ready()
+        # get ip and port for motorcad
+        address = self.pim_instance.services["http"].uri
+
+        ip = address.split(":")[0] + ":" + address.split(":")[1]
```
plule-ansys, line 262: This address split is potentially buggy and I believe is unnecessary, since in the end the request passed to `requests` is `address + "/jsonrpc"`. Moreover the gobal `SERVER_IP` will prevent to connect to multiple remote sessions and create unexpected bugs: in local we multiply sessions with many ports, but in a container environment we multiply sessions with many remote hosts.

I strongly suggest to make `url` a member of `MotorCADConnection` instead of just `ip`, which solves both issues: in `_launch_motorcad_local()`, `url` would take the value `"http://localhost:" + port + "/jsonrpc"`, in `_launch_motorcad_remote()`, `url` would take the value `pim_instance.services["http"].uri + "/jsonrpc"`. I think that even outside of PIM, like a custom scenario where a user manages the launching themselves through a VM, docker-compose or anything, they can simply provide the URL themselves without bothering about the whole ip/port separation.

jgsdavies: Thanks @plule-ansys . I've come across this issue before. I'm going to change this in another PR.

plule-ansys (dismissed):

src/ansys/motorcad/core/rpc_client_core.py
plule-ansys: @jgsdavies Ok! that was the only issue I saw in this PR

JackB-Ansys (approved):

JackB-Ansys (dismissed):

JackB-Ansys (approved):

PR #85, opened 2023-02-07
Graph issue 79

ravi-ansys:
Added get_graph which takes function arguments for automation of all graphs (temperature, magnetic, power)
added graph methods and respective tests for thermal transient temperature and power loss graphs


JackB-Ansys (approved):
Checked the feature and some tests, all looks great.

jgsdavies (approved):

PR #125, opened 2023-04-20
Try resolve localhost on connection

jgsdavies:


src/ansys/motorcad/core/rpc_client_core.py
```diff
@@ -14,7 +14,9 @@
 
 DEFAULT_INSTANCE = -1
 
-SERVER_IP = "http://localhost"
+LOCALHOST_ADDRESS = "http://localhost"
+TRY_RESOLVE_LOCALHOST = True
```
jgsdavies, line 18: Flag so we can enable/disable this behaviour. Potentially useful for testing/disabling if there are issues

src/ansys/motorcad/core/rpc_client_core.py
```diff
@@ -188,15 +193,19 @@ def __init__(
...
-
         # Check for response
         if self._wait_for_response(2) is True:
             self._connected = True
-            # Store Motor-CAD version number for any required backwards compatibility
 
+            # Store Motor-CAD version number for any required backwards compatibility
             self.program_version = self._get_program_version()
 
+            self.pid = self.get_process_id()
```
jgsdavies, line 203: Unrelated change but I think this was a mistake to not have this after the connection check

src/ansys/motorcad/core/rpc_client_core.py
```diff
@@ -205,6 +214,32 @@ def __init__(
...
             )
 
+    def _resolve_localhost(self):
+        global SERVER_IP
+
+        ipv6_localhost = "http://[::1]"
+        ipv4_localhost = "http://127.0.0.1"
+
+        if self._check_address_for_response(ipv6_localhost):
+            SERVER_IP = ipv6_localhost
```
jgsdavies, line 224: I can only manage to test ipv6 on my machine/in our unit tests but I have verified that ipv4 is correct in the past - I used this before I was using the new Ansys VPN. 

Also I think if both of these attempts are unsuccessful then it will revert to using "localhost". My main concern is that this will add a delay to the initial connection

src/ansys/motorcad/core/rpc_client_core.py
```diff
@@ -209,6 +224,32 @@ def __init__(
...
+        elif self._check_address_for_response(ipv4_localhost):
+            SERVER_IP = ipv4_localhost
+        else:
+            SERVER_IP = LOCALHOST_ADDRESS
+
+    def _check_address_for_response(self, address):
```
sarah-r-woodrow, line 240: Could you please add a docstring to summarise what this function does, and some comments to explain why we have to set SERVER_IP, how it is used and why we have to restore.

sarah-r-woodrow (changes requested):

src/ansys/motorcad/core/rpc_client_core.py
jgsdavies: I moved          

`self.pid = self.get_process_id()`

src/ansys/motorcad/core/rpc_client_core.py
jgsdavies: I think better to add later if needed.

src/ansys/motorcad/core/rpc_client_core.py
```diff
@@ -185,22 +192,30 @@
         if self.reuse_parallel_instances is True:
             self._open_new_instance = False
 
-        if _HAS_PIM and pypim.is_configured():
+        if self._url != "":
+            # Don't want to start a Motor-CAD instance
```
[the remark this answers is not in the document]
jgsdavies: No. I'm assuming since a full url is specified that the instance exists. I think in the future I might add open_new_motorcad() and connect_to_motorcad() functions which return a MotorCAD object. I think this class init function/args are starting to get a bit confusing.

src/ansys/motorcad/core/rpc_client_core.py
```diff
@@ -238,6 +242,11 @@ def _resolve_localhost(self):
             SERVER_IP = LOCALHOST_ADDRESS
 
     def _check_address_for_response(self, address):
+        """Try resolving localhost to a specified url.
+
+        We can use a handshake method with Motor-CAD instance to check if connection is successful.
+        """
+        # Need to use SERVER_IP global var to preserve backwards compatibility
```
sarah-r-woodrow, line 249: I'm not sure this really explains things clearly enough. Backwards compatibility with what? I thought we needed to use it as _wait_for_response communicates using the address from get_url, which in turn uses the global SERVER_IP

sarah-r-woodrow (changes requested):

src/ansys/motorcad/core/rpc_client_core.py
```diff
@@ -185,22 +192,30 @@ def __init__(
...
             self._connected = True
 
             # Store Motor-CAD version number for any required backwards compatibility
             self.program_version = self._get_program_version()
 
             self.pid = self.get_process_id()
 
+            if (SERVER_IP == LOCALHOST_ADDRESS) and TRY_RESOLVE_LOCALHOST:
+                # Try to resolve localhost to speed up function calls
+                self._resolve_localhost()
```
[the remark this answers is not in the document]
sarah-r-woodrow: Not covered directly as if this fails then all other tests will fail, so it will be obvious this is a problem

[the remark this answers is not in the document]
sarah-r-woodrow: As discussed, think it would be useful to test this independently. If the resolving fails then other tests may still pass but might be slower as they will just use the localhost address.

sarah-r-woodrow (changes requested):

src/ansys/motorcad/core/rpc_client_core.py
jgsdavies: This was in my initial commit but I think it might have been done in another PR too and merged from main so no longer showing.

src/ansys/motorcad/core/rpc_client_core.py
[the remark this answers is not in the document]
jgsdavies: open_new_instance is True by default. I don't want users to have to specify open_new_instance=False to avoid error if using url. I think separate launch functions are needed really. I've added as an issue: https://github.com/ansys/pymotorcad/issues/142

src/ansys/motorcad/core/motorcad_methods.py
```diff
@@ -17,13 +17,17 @@ def __init__(
...
+        url="",
+        timeout=2,
     ):
         self.connection = _MotorCADConnection(
             port,
             open_new_instance,
             enable_exceptions,
             enable_success_variable,
             reuse_parallel_instances,
+            url=url,
```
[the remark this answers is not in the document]
jgsdavies: I wanted this functionality in main for HPC work. In hindsight I think this was a bad idea to add as part of this PR. I have tidied it up a bit and added a test. Shouldn't be super critical as long as current tests pass - this functionality will be tested a lot more as part of HPC work

src/ansys/motorcad/core/rpc_client_core.py
jgsdavies: Maybe not super clear comment. I'd like to get rid of SERVER_IP ideally and have this as a class var. I've used it in a few places however so I'm going to keep it. I'm hoping url parameter will make this obsolete anyway. I'll improve the comments

src/ansys/motorcad/core/rpc_client_core.py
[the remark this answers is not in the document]
sarah-r-woodrow: Okay, cheers

src/ansys/motorcad/core/rpc_client_core.py
sarah-r-woodrow: Looks good, cheers

sarah-r-woodrow (approved):

PR #147, opened 2023-07-27
Add template list

ravi-ansys:
Created a function to return a list of templates with info. 
Function reads the templates under installation folder


jgsdavies (changes requested):
Will review after Jack B.

PR #150, opened 2023-08-01
fix: update ci_cd

Revathyvenugopal162:
- [x] Update the pyproject.toml with version rather than dynamic version 
- [x] Update the python version
- [x] Add package namespace


jgsdavies (approved):
Thanks for the help @Revathyvenugopal162!

PR #155, opened 2023-08-04
Testing/update tests for new motorcad

jgsdavies:


jgsdavies:
Updating Motor-CAD exe to test new geometry methods

matthewAnsys (approved):

PR #173, opened 2023-08-31
New unit tests for collision detection api

matthewAnsys:


src/ansys/motorcad/core/geometry.py
```diff
@@ -157,6 +159,21 @@ def is_closed(self):
         else:
             return False
 
+    def contains_same_entities(self, other):
```
jgsdavies, line 162: Does this need to exist or can we expand the other function:

```def entities_same(entities_a, entities_b, check_direction=False):```

File: src/ansys/motorcad/core/geometry.py [hunk 2]
```diff
@@ -516,6 +545,57 @@ def get_entities_have_common_coordinate(entity_1, entity_2):
         return False
 
 
+def entities_same(entities_a, entities_b):
```
jgsdavies, line 548: Need to add these to docs manually in: \doc\source\methods\geometry_functions.rst

jgsdavies (changes requested):

jgsdavies (approved):

PR #191, opened 2023-09-20
Adaptive templates delete region API call

matthewAnsys:


src/ansys/motorcad/core/methods/adaptive_geometry.py
```diff
@@ -160,3 +160,22 @@ def unite_regions(self, region, regions):
         region._from_json(united_raw)
 
         return region
+
+    def delete_region(self, region, remove_children=False):
+        """Delete region from Motor-CAd geometry engine.
```
jgsdavies, line 165:
```suggestion
        """Delete region from Motor-CAD geometry engine.
```

jgsdavies (approved):

PR #199, opened 2023-10-03
Adaptive templates subtract region api call

matthewAnsys:


tests/test_geometry.py
```diff
@@ -312,6 +326,20 @@ def test_region_contains_same_entities():
...
 
 
+def test_region_get_parent():
+    pocket = mc.get_region("rotor pocket")
+    expected_region = mc.get_region("rotor")
+
+    assert pocket.get_parent(mc) == expected_region
+
+
+def test_region_get_children():
```
jgsdavies, line 336: This test is failing with the new version. I have updated the file a bit so can you take a look at this? 
image

jgsdavies (changes requested):

tests/test_geometry.py
jgsdavies: Region names not unique. Is this a problem?

matthewAnsys: No, not an issue, test file is using compatibility settings for unique naming, test shows it is doing the correct thing so happy to leave as is

src/ansys/motorcad/core/geometry.py
```diff
@@ -157,6 +163,119 @@ def is_closed(self):
...
+        """Property for child names list.
+
+        Returns
+        ----------
+        list of string
+            list of child region names
+        """
+        return self._child_names
+
+    def get_children(self):
```
jgsdavies, line 177: Worth doing the update to proper getter/setter functions? https://www.geeksforgeeks.org/getter-and-setter-in-python/

src/ansys/motorcad/core/geometry.py
```diff
@@ -157,6 +163,119 @@ def is_closed(self):
...
+        """
+        if type(regions) is not list:
+            regions = [regions]
+
+        self._check_connection()
+        united_region = self.mc.unite_regions(self, regions)
+        self.update(united_region)
+
+    def collides(self, regions):
+        """Subtract region from self, returning any additional regions.
```
jgsdavies, line 235: Wrong docstring

src/ansys/motorcad/core/geometry.py
```diff
@@ -157,6 +163,119 @@ def is_closed(self):
...
+        self.entities = deepcopy(region.entities)
+        self.parent_name = region.parent_name
+        self._child_names = region.child_names
+
+    def _check_connection(self):
+        """Check mc connection for region."""
+        if self.mc is None:
+            raise Exception(
+                "A Motor-CAD connection is required for this function"
+                + ", Please set self.mc to a valid Motor-CAD instance"
```
jgsdavies, line 274: also need to update this string if self.mc changes

src/ansys/motorcad/core/geometry.py
```diff
@@ -157,6 +163,119 @@ def is_closed(self):
...
+            )
+        if self.mc.connection._wait_for_response(1) is False:
+            raise Exception("Failed mc connection")
```
jgsdavies, line 277: Maybe copy string above for this too

src/ansys/motorcad/core/geometry.py
```diff
@@ -384,7 +503,8 @@ def get_coordinate_from_distance(self, ref_coordinate, distance):
                 angle = atan2(ref_coordinate.y, ref_coordinate.x) - (distance / self.radius)
 
         return Coordinate(
-            self.centre.x + self.radius * cos(angle), self.centre.y + self.radius * sin(angle)
+            self.centre.x + abs(self.radius) * cos(angle),
```
jgsdavies, line 506: You can do:
```self.centre + Coordinate(*rt_to_xy(radius, angle))```

Comment (jgsdavies) [reply src/ansys/motorcad/core/geometry.py:506]: ```*``` is to pass the tuple output of ```rt_to_xy``` to ```Coordinate```

File: src/ansys/motorcad/core/methods/adaptive_geometry.py [hunk 7]
```diff
@@ -192,3 +192,36 @@ def delete_region(self, region, remove_children=False):
...
+
+    def subtract_region(self, region, region_subtract):
+        """Subtract a single region from another region.
+
+        Parameters
+        ----------
+        region : ansys.motorcad.core.geometry.Region
+            Motor-CAD region object
+
+        region_subtract : ansys.motorcad.core.geometry.Region
```
jgsdavies, line 204: Can you improve this docstring e.g. Motor-CAD region that is subtracted from region

tests/test_geometry.py
```diff
@@ -765,37 +778,27 @@ def test_check_collisions_3():
...
-        square.add_entity(entity)
-
-    t_p = []
-    t_l = []
-
-    t_p.append(geometry.Coordinate(1, 2.2))
-    t_p.append(geometry.Coordinate(2.2, 1))
-    t_p.append(geometry.Coordinate(4, 4))
+    # create and add line entities to region from their respective points
+    add_line_entities_from_points(square, points_square)
```
jgsdavies, line 791: Not sure I like this function. Would prefer to have something like ```new_lines = create_lines_from_points()``` and append this to EntityList. ```square.entities += new_lines``` or ```square.add_entities(new_lines)```

jgsdavies (changes requested):

tests/test_geometry.py
matthewAnsys: Issue fixed

jgsdavies (changes requested):

jgsdavies (approved):

PR #215, opened 2023-10-12
Fix motorcad instance not set from subtraction, collides and unite

matthewAnsys:


src/ansys/motorcad/core/geometry.py
```diff
@@ -226,6 +226,11 @@ def parent(self):
         self._check_connection()
         return self.motorcad_instance.get_region(self.parent_name)
 
+    @parent.setter
+    def parent(self, region):
```
jgsdavies, line 230: When does this get used?

jgsdavies (changes requested):

src/ansys/motorcad/core/geometry.py
jgsdavies: Test

matthewAnsys: Added test

jgsdavies (approved):

PR #216, opened 2023-10-16
Adaptive templates mirror functionality for Region/Entity/Line/Arc/Coordinate

matthewAnsys:


jgsdavies (changes requested):
```test_entity_mirror``` seems to be failing on the test server. Is this working locally when you run pytest for the whole file/all files?

File: src/ansys/motorcad/core/geometry.py [hunk 1]
```diff
@@ -512,6 +585,61 @@ def midpoint(self):
...
+            Line entity to mirror Line about
+
+        Returns
+        ----------
+        ansys.motorcad.core.geometry.Line
+        """
+        if isinstance(mirror_line, Line):
+            return Line(self.start.mirror(mirror_line), self.end.mirror(mirror_line))
+        else:
+            Exception("Line can only be mirrored about Line()")
```
jgsdavies, line 641: Not covered by tests - also think this needs a ```raise```?

src/ansys/motorcad/core/geometry.py
```diff
@@ -282,6 +282,35 @@ def collides(self, regions):
...
+        ansys.motorcad.core.geometry.Region
+        """
+        region = deepcopy(self)
+        region.entities.clear()
+        region.centroid = self.centroid.mirror(mirror_line)
+        region.region_coordinate = self.region_coordinate.mirror(mirror_line)
+        region._child_names = []
+
+        if unique_name:
+            region.name = region.name + "_Mirrored"
```
jgsdavies, line 307: Can we cover this in tests too?

src/ansys/motorcad/core/geometry.py
jgsdavies: Some others like this

jgsdavies (changes requested):

jgsdavies (approved):

jgsdavies:
> `test_entity_mirror` seems to be failing on the test server. Is this working locally when you run pytest for the whole file/all files?

test

src/ansys/motorcad/core/geometry.py
jgsdavies, line 1: test

PR #233, opened 2023-12-05
Add clear_messages API method

jgsdavies:


jgsdavies:
Revisit when next Motor-CAD service pack available

src/ansys/motorcad/core/methods/rpc_methods_ui.py
```diff
@@ -145,3 +145,10 @@ def set_3d_component_visibility(self, group_name, component_name, visibility):
         method = "Set3DComponentVisibility"
         params = [group_name, component_name, visibility]
         return self.connection.send_and_receive(method, params)
+
+    def clear_messages(self):
+        """Clear messages in message display window."""
```
PipKat, line 150:
```suggestion
        """Clear messages in the message display window."""
```
Use an article if it helps with readability.

JackB-Ansys (approved):

PR #235, opened 2023-12-06
Extra API methods for Adaptive scripting improvements

jgsdavies:


src/ansys/motorcad/core/rpc_client_core.py
```diff
@@ -38,14 +38,21 @@
 DONT_CHECK_MOTORCAD_VERSION = False
 
 
+def _is_running_in_internal_scripting():
```
jgsdavies, line 41: Used so that reset_adaptive_geometry and draw_regions_debug do nothing in internal adaptive script.

matthewAnsys (dismissed):

PR #245, opened 2024-01-02
Drawing regions from list of points

jgsdavies:


src/ansys/motorcad/core/geometry.py
```diff
@@ -39,6 +40,14 @@ def __eq__(self, other):
             and self.entities == other.entities
         )
 
+    @classmethod
+    def from_coordinate_list(cls):
```
jgsdavies, line 44: Need to refactor geometry.py before this is possible. Think this needs doing anyway since this is getting huge

src/ansys/motorcad/core/geometry.py
```diff
@@ -526,8 +575,8 @@ class Entity(object):
 
     def __init__(self, start, end):
         """Initialise Entity."""
-        self.start = start
-        self.end = end
+        self.start = deepcopy(start)
```
jgsdavies, line 578: Trying to avoid people accidentally editing entities if they mutate a coordinate

@matthewAnsys thoughts?

src/ansys/motorcad/core/geometry.py
```diff
@@ -728,6 +799,66 @@ def coordinate_on_entity(self, coordinate):
...
+            Angle of rotation in degrees.
+
+        Returns
+        -------
+        Line
+        """
+        return Line(self.start.rotate(centre_point, angle), self.end.rotate(centre_point, angle))
+
+    def _line_equation(self):
+        """Get value of coefficients of line in format A*x + B*y = C."""
```
jgsdavies, line 819: Useful because  A inf/undefined for vertical lines when using (y = Ax + B)

src/ansys/motorcad/core/geometry.py
```diff
@@ -1162,3 +1351,37 @@ def rt_to_xy(radius, theta):
     y = coordinates_complex.imag
 
     return x, y
+
+
+class _Orientation(Enum):
```
jgsdavies, line 1356: Kept these as internal functions for now

src/ansys/motorcad/core/geometry.py
```diff
@@ -509,7 +518,47 @@ def mirror(self, mirror_line):
...
+        """
+        angle_r = radians(angle)
+        x_new = centre_point.x + (
+            (cos(angle_r) * (self.x - centre_point.x)) + (-sin(angle_r) * (self.y - centre_point.y))
+        )
+        y_new = centre_point.y + (
+            (sin(angle_r) * (self.x - centre_point.x)) + (cos(angle_r) * (self.y - centre_point.y))
+        )
+
+        return Coordinate(x_new, y_new)
```
jgsdavies, line 545: @matthewAnsys 

Decided to return new coordinate instead of mutating existing.

Not sure I like this but I think mutating coordinate is likely to confuse beginner Python users

src/ansys/motorcad/core/geometry_drawing.py
jgsdavies, line 1: This has been done in https://github.com/ansys/pymotorcad/pull/246

and should disappear from this PR when this other PR has been merged. No need to review this file here

src/ansys/motorcad/core/geometry_fitting.py
```diff
@@ -0,0 +1,213 @@
...
+
+from ansys.motorcad.core.geometry import Arc, Coordinate, EntityList, Line
+
+
+class _EntityType(Enum):
+    line = 0
+    arc = 1
+
+
+class _TestEntity:
```
jgsdavies, line 14: Used to check if list of points is within an acceptable tolerance for an entity

tests/test_geometry_drawing.py
jgsdavies, line 1: Will be reviewed as part of https://github.com/ansys/pymotorcad/pull/246

tests/test_geometry_fitting.py
```diff
@@ -0,0 +1,43 @@
+from ansys.motorcad.core.geometry import Arc, Coordinate, Line
+from ansys.motorcad.core.geometry_fitting import return_entity_list
+
+
+def test_return_entity_list():
```
jgsdavies, line 5: basic test - will almost certainly need to add to this to catch edge cases

src/ansys/motorcad/core/geometry_fitting.py
```diff
@@ -0,0 +1,200 @@
+"""Unit containing functions for drawing geometry from a list of points."""
```
matthewAnsys, line 1:
```suggestion
"""Unit containing functions for creating geometry using from a list of points."""
```

src/ansys/motorcad/core/geometry_fitting.py
```diff
@@ -0,0 +1,200 @@
...
+
+from copy import copy
+from enum import Enum
+
+from ansys.motorcad.core.geometry import Arc, Coordinate, EntityList, Line
+
+
+class _EntityType(Enum):
```
matthewAnsys, line 9: missing docstring

src/ansys/motorcad/core/geometry_fitting.py
matthewAnsys: can you add a docstring

src/ansys/motorcad/core/geometry_fitting.py
```diff
@@ -0,0 +1,200 @@
...
+    line = 0
+    arc = 1
+
+
+class _TestEntity:
```
matthewAnsys, line 14: Not sure I partically like this class name, but unsure of what it could be updated to

matthewAnsys (changes requested):

src/ansys/motorcad/core/geometry.py
jgsdavies: I personally agree but trying to avoid confusion with beginners.

```python

# mutating correct way
p1 = Coordinate()
p2 = copy(p1)
p2.rotate(90)

# suspect most people will try
p1 = Coordinate()
p2 = p1.rotate(90)
# p1 is p2 (confusing?)

# vs

p1 = Coordinate()
p2 = p1.rotate(90)

# p1 and p2 are different objects - works as expected
```

src/ansys/motorcad/core/geometry_fitting.py
jgsdavies: Me neither - open to suggestions :)

src/ansys/motorcad/core/geometry_fitting.py
jgsdavies: See other comment on private methods - added a short one anyway

src/ansys/motorcad/core/geometry.py
```diff
@@ -509,7 +540,47 @@ def mirror(self, mirror_line):
...
+
+    @classmethod
+    def from_polar_coords(cls, radius, theta):
```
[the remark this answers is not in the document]
jgsdavies: This is returning a Coordinate() object from polar coords. Basically an overloaded init. I think this name makes sense

src/ansys/motorcad/core/geometry.py
```diff
@@ -509,7 +540,47 @@ def mirror(self, mirror_line):
...
+        """Create Motor-CAD coordinate from polar coordinates.
```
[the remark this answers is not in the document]
jgsdavies: made this a bit clearer

matthewAnsys (approved):

PR #261, opened 2024-02-08
Rotate translate regions and fixes

jgsdavies:
Arc currently need start,end,centre,radius. On first inspection it seems that only centre or radius is required to fully define the arc however this is not the case:

2 valid solutions for start/end/centre
image

These are both valid solutions for arc with start/end/rad
image

We have some options:
1. force user to specify direction of arc
2. Assume arc smaller than 180 degrees

I've tried to implement idea 2 whilst still allowing user to force arc with greater than 180 deg by using all 4 params.

Nice side effect of these changes is that centre coord can become a property rather than a class object. This means rotation/translation can be handled by parent Entity class in same way for both arc and Line.

Hoping this makes the most sense as rounding corners etc is very simple when you only have to provide start/end/radius


src/ansys/motorcad/core/geometry.py
```diff
@@ -554,7 +552,11 @@ def rotate(self, centre_point, angle):
         centre_point : Coordinate
             point to rotate Coordinate around.
```
PipKat, line 553:
```suggestion
            Point to rotate coordinate around.
```

src/ansys/motorcad/core/geometry.py
```diff
@@ -435,13 +409,37 @@ def edit_point(self, old_coordinates, new_coordinates):
...
             if entity.end == old_coordinates:
                 entity.end = deepcopy(new_coordinates)
-            if isinstance(entity, Arc):
-                # Check Arc is still valid
-                _ = entity.centre
+                edited = True
+
+            if edited and isinstance(entity, Arc):
+                # Draw line between arc start/end
+                # get centre point of that line
```
PipKat, line 423:
```suggestion
                # Get center point of that line
```

src/ansys/motorcad/core/geometry.py
```diff
@@ -897,14 +876,29 @@ def get_line_intersection(self, line):
             return None
 
 
-class _BaseArc(Entity):
-    """Internal class to allow creation of Arcs."""
+class Arc(Entity):
+    """Python representation of Motor-CAD arc entity based upon start, end, centre and radius.
```
PipKat, line 880:
```suggestion
    """Python representation of Motor-CAD arc entity based upon start, end, centre, and radius.
```

JackB-Ansys (approved):

src/ansys/motorcad/core/geometry.py
jgsdavies: Motor-CAD API methods are British English so trying to have some consistency. Most methods are pre PyAnsys so decided it's too big a change for customers to switch now

src/ansys/motorcad/core/geometry.py
jgsdavies: Coordinate is a class name so keeping this capitalised

jgsdavies:
```suggestion
            Point to rotate Coordinate around.
```

src/ansys/motorcad/core/geometry.py
```diff
@@ -552,7 +552,7 @@ def rotate(self, centre_point, angle):
         Parameters
         ----------
         centre_point : Coordinate
-            point to rotate Coordinate around.
+            Point to rotate Coordinate around.
```
PipKat, line 555:
```suggestion
            Point to rotate coordinate around.
```
Only capitalize proper nouns or the first letter of the initial word (as you did).

JackB-Ansys (approved):

JackB-Ansys (approved):

PR #262, opened 2024-02-09
Add new option when connecting to Motor-CAD to keep_instance_open

JackB-Ansys:
Once the instance is done with (script finished) we can choose to keep the instance open. Previously had to use reuse_parallel_instances for this option, but that has additional functionality which could cause problems.

### DIFF
--- a/src/ansys/motorcad/core/rpc_client_core.py
+++ b/src/ansys/motorcad/core/rpc_client_core.py
@@ -156,6 +156,7 @@ class _MotorCADConnection:
         enable_exceptions,
         enable_success_variable,
         reuse_parallel_instances,
+        keep_instance_open,
         url="",
         timeout=2,
         compatibility_mode=False,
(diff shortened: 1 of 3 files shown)


ravi-ansys (dismissed):

src/ansys/motorcad/core/rpc_client_core.py
```diff
@@ -306,7 +311,11 @@ def _close_motorcad_on_exit(self):
             and (self._compatibility_mode is False)
         ):
             # Local Motor-CAD has been launched by Python
-            return True
+            if self.keep_instance_open is True:
```
gmalinve, line 314:
```suggestion
            if self.keep_instance_open:
```

ravi-ansys (approved):

jgsdavies:
@JackB-Ansys Looks good. Do you want this available immediately in a new minor release?

jgsdavies (approved):

JackB-Ansys:
> 

Maybe the same release as Adaptive Templates Examples?

PR #314, opened 2024-05-31
Fix broken test when run in parallel

jgsdavies:


tests/test_rpc_client_core.py
```diff
@@ -61,7 +62,15 @@ def test_internal_connection(mc):
 def test_open_new_with_port():
     test_port = 36020
 
-    mc2 = MotorCAD(open_new_instance=True, port=test_port)
+    try:
```
jgsdavies, line 65: This doesn't happen very often so happy to just skip the test if it occurs

jgsdavies:
See test run https://github.com/ansys/pymotorcad/actions/runs/9312732168/job/25633969172 for example

JackB-Ansys (approved):

PR #332, opened 2024-07-05
Fix get_coordinate_from_distance() for Arc

JackB-Ansys:
Fixed get_coordinate_from_distance() _BaseArc class method so that correct point on the arc is always returned.

Fixes a bug raised by customer.


src/ansys/motorcad/core/geometry.py
```diff
@@ -1118,20 +1118,17 @@ def get_coordinate_from_distance(self, ref_coordinate, distance):
...
+        ref_coordinate_angle = atan2(
+            (ref_coordinate.y - self.centre.y), (ref_coordinate.x - self.centre.x)
+        )
         if ref_coordinate == self.end:
-            if self.radius >= 0:
-                # anticlockwise
-                angle = atan2(ref_coordinate.y, ref_coordinate.x) - (distance / self.radius)
-            else:
-                angle = atan2(ref_coordinate.y, ref_coordinate.x) + (distance / self.radius)
+            e = -1
```
jgsdavies, line 1125: Update unit test to cover this

jgsdavies (changes requested):

matthewAnsys (dismissed):

src/ansys/motorcad/core/geometry.py
```diff
@@ -1164,19 +1164,7 @@ def length(self):
...
-            arc_angle = 0
-        elif ((self.radius > 0) and (angle_1 > angle_2)) or (
-            (self.radius < 0) and angle_2 < angle_1
-        ):
-            arc_angle = angle_2 - (angle_1 - 360)
-        else:
-            arc_angle = angle_2 - angle_1
-
-        return self.radius * radians(arc_angle)
+        return abs(self.radius * radians(self.total_angle))
```
jgsdavies, line 1164: This is in another PR too? 
https://github.com/ansys/pymotorcad/pull/338
We shouldn't commit this in this PR. Either remove manually or wait until #338 is in main and then merge and resolve conflicts

jgsdavies (changes requested):

src/ansys/motorcad/core/geometry.py
JackB-Ansys: Yes I think I found this issue while fixing the other one. I think best to wait until 338 is in main then merge this one

jgsdavies (approved):

PR #335, opened 2024-07-08
Add Region class method for Corner Rounding

JackB-Ansys:
A method for defining automatic corner rounding for regions. 
The method finds entities that contain the specified corner coordinate, cuts the entities by a certain distance and draws an arc between the new start/end coordinates.

Example shown for triangular rotor notch example, .mot file for this example attached (in zip):
image
[e9_Tri_Rotor_Notch_Rounded.zip](https://github.com/user-attachments/files/16132080/e9_Tri_Rotor_Notch_Rounded.zip)
Trapezoidal Duct example with corner rounding also attached (in zip): 
[Trapezoidal_duct_rounded.zip](https://github.com/user-attachments/files/16141696/Trapezoidal_duct_rounded.zip)


jgsdavies (changes requested):
Spotted some cases that need handling:

1. Rounding a corner with arc/line intersection doesn't quite seem right:
image

This is a slightly more complicated case. Happy to also just throw an exception for now if you don't want to implement this. Also need to consider arc/arc intersection.

2. I think we should check that the intersecting lines have the length the perform the requested rounding. At the moment I can increase the radius as far as I like with no errors
image

src/ansys/motorcad/core/geometry.py
```diff
@@ -493,6 +493,10 @@ def edit_point(self, old_coordinates, new_coordinates):
     def round_corner(self, corner_coordinate, radius):
         """Round the corner of a region.
 
+        The corner coordinate must already exist on two entities belonging to the region.
+        The two entities adjacent to the corner will be shortened and an arc will be created between
+        them.
```
PipKat, line 498:
```suggestion
        The corner coordinates must already exist on two entities belonging to the region.
        The two entities adjacent to the corner are shortened, and an arc is created between
        them.
```
Use present tense.

JackB-Ansys:
> Spotted some cases that need handling:
> 
> 1. Rounding a corner with arc/line intersection doesn't quite seem right:
>    image
> 
> This is a slightly more complicated case. Happy to also just throw an exception for now if you don't want to implement this. Also need to consider arc/arc intersection.
> 
> 2. I think we should check that the intersecting lines have the length the perform the requested rounding. At the moment I can increase the radius as far as I like with no errors
>    image

I think we can avoid these issues with the new commits. Now an exception is thrown if a corner radius is given that would shorten an entity by more than its length

src/ansys/motorcad/core/geometry.py
```diff
@@ -520,11 +524,11 @@ def round_corner(self, corner_coordinate, radius):
                 "Point found on only one entity in region. "
                 "You must specify a corner in this region."
             )
-        # if first and last entities, swap the positions
+        # if the adjacent entities are the first and last entities of the region, swap the positions
         if entity_indicies[0] == 0 and entity_indicies[1] == len(self.entities) - 1:
             entity_indicies[0], entity_indicies[1] = entity_indicies[1], entity_indicies[0]
             adj_entities[0], adj_entities[1] = adj_entities[1], adj_entities[0]
-        # Calculate distances for adjacent entities
+        # Calculate distances by which the adjacent entities will be shortened
```
PipKat, line 531:
```suggestion
        # Calculate distances by which the adjacent entities are shortened
```

src/ansys/motorcad/core/geometry.py
```diff
@@ -561,17 +565,43 @@ def round_corner(self, corner_coordinate, radius):
         half_chord = radius * sin(radians(arc_angle) / 2)
         distance = abs(half_chord / (sin(radians(corner_angle) / 2)))
 
+        # check that distances by which the adjacent entities will be shortened are less than  the
```
PipKat, line 568:
```suggestion
        # check that the  distances by which the adjacent entities are shortened are less than the
```

src/ansys/motorcad/core/geometry.py
```diff
@@ -561,17 +565,43 @@ def round_corner(self, corner_coordinate, radius):
...
+        # lengths of the adjacent entities.
         for index in range(len(adj_entities)):
             j = adj_entities[index]
+            if j.length < distance:
+                raise Exception(
+                    "Corner radius too large for these entities. "
```
PipKat, line 574:
```suggestion
                    "Corner radius is too large for these entities. "
```
The string should have a verb.

src/ansys/motorcad/core/geometry.py
```diff
@@ -561,17 +565,43 @@ def round_corner(self, corner_coordinate, radius):
...
+        # create the new corner arc and insert it between the adjacent entities
         corner_arc = Arc(coordinates[0], coordinates[1], radius=radius)
         self.insert_entity(entity_indicies[0] + 1, corner_arc)
 
+    def round_corners(self, corner_coordinates, radius):
+        """Round multiple corners of a region.
+
+        Each corner coordinate must already exist on two entities belonging to the region.
+        The two entities adjacent to each corner will be shortened and an arc will be created
+        between them.
```
PipKat, line 593:
```suggestion
        The two entities adjacent to each corner are shortened, and an arc is created
        between them.
```

src/ansys/motorcad/core/geometry.py
```diff
@@ -561,17 +565,43 @@ def round_corner(self, corner_coordinate, radius):
...
+
+        Parameters
+        ----------
+        corner_coordinates : list of ansys.motorcad.core.geometry.Coordinate
+            List of coordinates of the corners to be rounded.
+        radius : float
+            Radius by which the corners will be rounded.
```
PipKat, line 600:
```suggestion
            Radius to round the corners by.
```

JackB-Ansys:
I think all the necessary tests should now be there.

src/ansys/motorcad/core/geometry.py
```diff
@@ -515,6 +515,140 @@ def edit_point(self, old_coordinates, new_coordinates):
...
+        The corner coordinates must already exist on two entities belonging to the region.
+        The two entities adjacent to the corner are shortened, and an arc is created between
+        them.
+
+        Parameters
+        ----------
+        corner_coordinate : ansys.motorcad.core.geometry.Coordinate
+            Coordinate of the corner to be rounded.
+        radius : float
+            Radius by which the corner will be rounded.
```
PipKat, line 530:
```suggestion
        corner_coordinate : ansys.motorcad.core.geometry.Coordinate
            Coordinate of the corner to round.
        radius : float
            Radius to round the corner by.
```
Use present tense rather than future tense verbs.

src/ansys/motorcad/core/geometry.py
```diff
@@ -515,6 +515,140 @@ def edit_point(self, old_coordinates, new_coordinates):
...
+        """
+        # If radius is 0, do nothing
+        if radius == 0:
+            return
+
+        # find adjacent entities. There should be 2 entities adjacent to the corner. Going
```
PipKat, line 536: When a comment consists of multiple sentences, it is preferable, but not mandatory, to start them all with a capital letter.

src/ansys/motorcad/core/geometry.py
```diff
@@ -515,6 +515,140 @@ def edit_point(self, old_coordinates, new_coordinates):
...
+        """Round multiple corners of a region.
+
+        Each corner coordinate must already exist on two entities belonging to the region.
+        The two entities adjacent to each corner are shortened, and an arc is created
+        between them.
+
+        Parameters
+        ----------
+        corner_coordinates : list of ansys.motorcad.core.geometry.Coordinate
+            List of coordinates of the corners to be rounded.
```
PipKat, line 645:
```suggestion
            List of coordinates of the corners to round.
```

ravi-ansys:
image
May be a tolerance is required as some user may try to specify the point as shown in the GUI

james-packer (approved):
@jgsdavies, I've reviewed this, and made a change for the arc case (to improve the intersection calculation for large radius rounding on arcs). Please could you check that my changes are OK?

jgsdavies (approved):

PR #338, opened 2024-07-15
Corrected calculation of arc length

JackB-Ansys:
previous arc length calculation had problems, replaced with a simpler calculation that relies of the total_angle() class method/property of arcs


jgsdavies (changes requested):
Can you improve the unit test for this function by adding something that would have failed previously.

JackB-Ansys:
> Can you improve the unit test for this function by adding something that would have failed previously.

New test added

jgsdavies (approved):

PR #343, opened 2024-07-19
added-thermal-twinbuilder-rom-example

HusainAdam-Ansys:
### Created new example for the Twin Builder Motor-CAD ROM component. 

The Twin Builder Motor-CAD ROM component is a new component available in Twin Builder 2024 R2. A set of Motor-CAD output files must be provided as an input to create this component (the creation step is done directly within Twin Builder).

This example script is an example showcasing how this can be done, ensuring the correct Motor-CAD operations are performed the correct number of times, with the Motor-CAD output files then formatted in the appropriate manner. 

Note: This script uses a Class to define the functionality. It's not possible to add code comments within the class definition, hence detailed code comments have been left. 

(I did make an attempt to remove the class structure but this leads to more complexity and lots of globals, so I reverted this).


jgsdavies (changes requested):
Can you make sure that only files that are needed to build the example are added to git. Not familiar with this workflow but I can see some files that should not be tracked.

JackB-Ansys (approved):

PR #376, opened 2024-09-06
Add custom loss functions. 

ShreyaGursal:
Add custom loss functions for internal/external.


shreya_custom_loss_test.py
```diff
@@ -0,0 +1,73 @@
...
+# from numpy import error_message
+
+from ansys.motorcad.core import MotorCAD
+
+mc = MotorCAD(open_new_instance=False)
+
+# mc.do_magnetic_calculation()
+
+def add_internal_custom_loss(name, function, type, thermal_node):
+    if type not in ["Electrical", "Mechanical"]:
```
jgsdavies, line 13: Think about case sensitivity. What happens if the user enters "electrical"?

shreya_custom_loss_test.py
```diff
@@ -0,0 +1,73 @@
...
+        mc.set_array_variable("CustomLoss_ThermalNode_Internal_Lab", no_internal_losses, thermal_node)
+
+def add_external_custom_loss(name, power_function, voltage_function):
+    no_external_losses = mc.get_variable("NumCustomLossesExternal_Lab")
+    mc.set_variable("NumCustomLossesExternal_Lab", no_external_losses + 1)
+    mc.set_array_variable('CustomLoss_Name_External_Lab', no_external_losses, name)
+    mc.set_array_variable('CustomLoss_PowerFunction_External_Lab', no_external_losses, power_function)
+    mc.set_array_variable('CustomLoss_VoltageFunction_External_Lab', no_external_losses, voltage_function)
+
+def remove_internal_custom_loss(index):
```
jgsdavies, line 33: Suggestion to reduce code duplication. Something like this?

Left the body empty so you can try and finish it.

```python
# preceding underscore makes it clear to user that this is a function they shouldn't be using ( private method
# https://www.datacamp.com/tutorial/python-private-methods-explained
def _motorcad_array_pop(index, var_length_array, list_of_var_names):
    no__losses = mc.get_variable(var_length_array)

    # Loop through each item in list_of_var_names and remove from array

    mc.set_variable(var_length_array, no__losses - 1)

# useage example
# [] is a list which we can iterate through and can be different lengths
motorcad_array_pop(1,
                   "NumCustomLossesExternal_Lab",
                   ["CustomLoss_Name_External_Lab","CustomLoss_PowerFunction_External_Lab", "CustomLoss_VoltageFunction_External_Lab"])
```

shreya_custom_loss_test.py
```diff
@@ -0,0 +1,73 @@
...
+def add_internal_custom_loss(name, function, type, thermal_node):
+    if type not in ["Electrical", "Mechanical"]:
+        raise ValueError("Thermal Loss Type must be Electrical or Mechanical")
+    else:
+        # Find number of losses
+        no_internal_losses = mc.get_variable("NumCustomLossesInternal_Lab")
+        # Increase array length by 1
+        mc.set_variable("NumCustomLossesInternal_Lab", no_internal_losses + 1)
+        # Define properties
+        mc.set_array_variable("CustomLoss_name_internal_lab", no_internal_losses, name)
```
jgsdavies, line 21: When we have strings we are repeatedly using I would suggest making a global constant that we can use instead

```python
kCustomLoss_name_internal_lab = "CustomLoss_name_internal_lab"
```

jgsdavies (changes requested):

shreya_custom_loss_test.py
```diff
@@ -0,0 +1,74 @@
...
+kCustomLoss_Type_Internal_Lab = "CustomLoss_Type_Internal_Lab"
+kCustomLoss_ThermalNode_Internal_Lab = "CustomLoss_ThermalNode_Internal_Lab"
+kNumCustomLossesExternal_Lab = "NumCustomLossesExternal_Lab"
+kCustomLoss_Name_External_Lab = "CustomLoss_Name_External_Lab"
+kCustomLoss_PowerFunction_External_Lab = "CustomLoss_PowerFunction_External_Lab"
+kCustomLoss_VoltageFunction_External_Lab = "CustomLoss_VoltageFunction_External_Lab"
+
+def add_internal_custom_loss(name, function, type, thermal_node):
+    type = type.capitalize()
+    if type not in ["Electrical", "Mechanical"]:
```
jgsdavies, line 23: This will work but I prefect removing
```
 type = type.capitalize()
```

and using

```suggestion
    if type.lower() not in ["electrical", "mechanical"]:
```

I think this is more robust

shreya_custom_loss_test.py
```diff
@@ -0,0 +1,74 @@
...
+
+def add_external_custom_loss(name, power_function, voltage_function):
+    no_external_losses = mc.get_variable(kNumCustomLossesExternal_Lab)
+    mc.set_variable(kNumCustomLossesExternal_Lab, no_external_losses + 1)
+    mc.set_array_variable(kCustomLoss_Name_External_Lab, no_external_losses, name)
+    mc.set_array_variable(kCustomLoss_PowerFunction_External_Lab, no_external_losses, power_function)
+    mc.set_array_variable(kCustomLoss_VoltageFunction_External_Lab, no_external_losses, voltage_function)
+
+def remove_internal_custom_loss(name):
+    no_internal_losses = mc.get_variable(kNumCustomLossesInternal_Lab)
```
jgsdavies, line 46: What about a function like this:

```python
def _get_index_from_name(custom_loss_name, array_name):
    # Get index from name
    return index
```

jgsdavies (changes requested):

shreya_custom_loss_test.py
jgsdavies: As discussed  -leave a comment to explain this

src/ansys/motorcad/core/methods/rpc_methods_lab.py
```diff
@@ -113,3 +124,150 @@ def calculate_duty_cycle_lab(self):
...
+        """Add an internal custom loss.
+
+        Parameters
+        ----------
+        name : str
+            Name of lab internal custom loss
+        function : str
+            Function of lab internal custom loss
+        type : str
+            Type of lab internal custom loss. Options are Electrical or Mechanical
```
jgsdavies, line 138:
```suggestion
            Type of lab internal custom loss. Options are ``Electrical`` or ``Mechanical``
```

Makes the option more clear. e.g.
image

jgsdavies (changes requested):

jgsdavies (changes requested):

PR #386, opened 2024-09-24
Fix max recursion depth reached with geometry drawing

jgsdavies:
Issue caused by algorithm trying the same point repeatedly 

@JackB-Ansys Can you test this with your file?


JackB-Ansys (approved):

JackB-Ansys:
> Issue caused by algorithm trying the same point repeatedly
> 
> @JackB-Ansys Can you test this with your file?

Checked with e1 template example that previously triggered an error. Now successfully draws geometry with appropriate labels.

PR #388, opened 2024-09-26
Lab Model Export API

ShreyaGursal:
File -> Export Lab model, to be triggerable via a python command for use with MotorCAD, optiSLang, and Concept EV.
MotorCAD PR: https://github.com/ansys-internal/motor-cad/pull/266


tests/test_lab.py
```diff
@@ -166,3 +171,32 @@ def test_external_custom_loss_functions(mc):
...
+def test_lab_model_export():
+    mc = MotorCAD(open_new_instance=False)
+    mc.set_variable("MessageDisplayState", 2)
+    mc.load_template("e8")
+    file_path = get_dir_path() + r"\test_files\temp_files\lab_model_export.lab"
+
+    if path.exists(file_path):
+        remove(file_path)
+
+    assert path.exists(file_path) is False
```
jgsdavies, line 185: Let's also check some failure cases in this function. e.g. lab model file not built

We can also try and force one of these other cases. e.g. change lab model build speed:
image

and check the error message contains some expected keywords

image

jgsdavies (changes requested):

jgsdavies (approved):

PR #411, opened 2024-10-23
Set solid/laminated material type for adaptive regions

jgsdavies:
Motor-CAD PR https://github.com/ansys-internal/motor-cad/pull/500


james-packer:
[DefaultMotor_WithStatorBridge_TestLaminationType.zip](https://github.com/user-attachments/files/17598838/DefaultMotor_WithStatorBridge_TestLaminationType.zip)
Tested that lamination type follows the default setting if not set, and that the lamination type can be overridden to either laminated or solid, and Je is non-zero if solid, and zero if laminated, all works as expected.

james-packer (dismissed):
Tested locally, and confirmed tests pass locally (fail on automated run because Motor-CAD is not using this branch of PyMotorCAD on test server).

james-packer (approved):
Re-ran automated tests and manual tests that an adaptive region in the stator can be set to default (and follows stator lamination type), or correctly overrides to laminated or solid.

james-packer (approved):
After resolving simple merge conflicts with linked regions.

PR #422, opened 2024-11-01
Add new class methods for returning intersection coordinates of entities

JackB-Ansys:
We don't have a method for finding:

Intersection of a Line and Arc

Intersection of Arc and another Arc

(only Line and Line)

This would be a useful method.


JackB-Ansys:
Need to add tests

james-packer (approved):
Reviewed. Added special case handling for vertical line intersection with arc. @JackB-Ansys has reviewed my commits.

PR #427, opened 2024-11-08
Fix graph tests following change to motor test file in #411.

james-packer:
Update values used in tests, and relax test tolerance for graphs tests.

### DIFF
--- a/tests/test_graphs.py
+++ b/tests/test_graphs.py
@@ -111,12 +111,15 @@ def test_get_heatflow_graph(mc):
     mc.set_variable("TransientCalculationType", 0)
     mc.do_transient_analysis()
 
-    x, y = mc.get_heatflow_graph("Housing OH [Rear] to Ambient")
+    x, y = mc.get_heatflow_graph("TVent Airgap to Rotor Surface")
+    # This should be the same grap series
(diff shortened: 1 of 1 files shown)


JackB-Ansys (dismissed):

tests/test_graphs.py
```diff
@@ -116,7 +116,7 @@ def test_get_heatflow_graph(mc):
 
     assert len(x) == len(y) == 9
     assert len(x1) == len(y1) == 11
-    assert almost_equal(y1[3], 11.68)
+    assert almost_equal(y1[3], 12.45)
```
JackB-Ansys, line 119: I am thinking we could make this more generic and do something like just check y1[3] > y1[4]? My concern is that the results may change due to calculation setting changes in future releases of the software, and we don't want to have to keep updating the values.

james-packer: Agreed, I've done this and commented about why this makes sense for this series.

james-packer: I have also changed the other tests so they use the equivalent graph_point functions, and cross check the results. This isn't possible in this case, as there isn't a get_heatflow_graph_point function.

JackB-Ansys (approved):

PR #443, opened 2024-12-13
[feat] Update json region conversion for base name from API update

matthewAnsys:
Update json region conversion for base name from API update


src/ansys/motorcad/core/geometry.py
```diff
@@ -93,7 +93,7 @@ class RegionType(Enum):
 
 class Region(object):
     """Create geometry region.
-
+
```
matthewAnsys, line 96:
```suggestion
```

matthewAnsys:
This PR fixes the issue with the i10 template not loading correctly due to unique names.

PR #447, opened 2024-12-18
[feat] Add linked regions api 

matthewAnsys:
Add adaptive templates linked regions api

### DIFF
--- a/src/ansys/motorcad/core/geometry.py
+++ b/src/ansys/motorcad/core/geometry.py
@@ -48,7 +48,7 @@ class Region(object):
         self._motorcad_instance = motorcad_instance
         self._region_type = RegionType.adaptive
         self.mesh_length = 0
-        self._linked_region = None
+        self._linked_regions = []
         self._singular = False
         self._lamination_type = ""
 
@@ -199,6 +199,9 @@ class Region(object):
         if "lamination_type" in json:
             new_region._lamination_type = json["lamination_type"]
 
+        if "linked_regions" in json:
+            new_region._linked_regions = json["linked_regions"]
+
         return new_region
 
     # method to convert python object to send to Motor-CAD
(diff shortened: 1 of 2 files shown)


src/ansys/motorcad/core/geometry.py
```diff
@@ -312,8 +316,9 @@ def _to_json(self):
             "entities": _convert_entities_to_json(self.entities),
             "parent_name": self._parent_name,
             "region_type": self._region_type.value,
-            "mesh_length": self._mesh_length,
-            "on_boundary": False if self._linked_region is None else True,
+            "mesh_length": self.mesh_length,
+            "linked_regions": self._linked_regions,
+            "on_boundary": False if len(self._linked_regions) is None else True,
```
matthewAnsys, line 321:
```suggestion
            "on_boundary": False if len(self._linked_regions) > 0 else True,
```

src/ansys/motorcad/core/geometry.py
```diff
@@ -312,8 +316,9 @@ def _to_json(self):
             "entities": _convert_entities_to_json(self.entities),
             "parent_name": self._parent_name,
             "region_type": self._region_type.value,
-            "mesh_length": self._mesh_length,
-            "on_boundary": False if self._linked_region is None else True,
+            "mesh_length": self.mesh_length,
+            "linked_regions": self._linked_regions,
+            "on_boundary": False if len(self._linked_regions) > 0 else True,
```
matthewAnsys, line 321:
```suggestion
            "on_boundary": False if len(self._linked_regions) = 0 else True,
```

src/ansys/motorcad/core/geometry.py
```diff
@@ -312,8 +316,9 @@ def _to_json(self):
             "entities": _convert_entities_to_json(self.entities),
             "parent_name": self._parent_name,
             "region_type": self._region_type.value,
-            "mesh_length": self._mesh_length,
-            "on_boundary": False if self._linked_region is None else True,
+            "mesh_length": self.mesh_length,
+            "linked_regions": self._linked_regions,
+            "on_boundary": False if len(self._linked_regions) = 0 else True,
```
matthewAnsys, line 321:
```suggestion
            "on_boundary": False if len(self._linked_regions) == 0 else True,
```

PR #465, opened 2025-01-30
Adaptive Templates Corner Rounding Example - e10 template different corner radii

JackB-Ansys:
image

### DIFF
--- a/examples/adaptive_library/IPMRoundRotorLamMagnets.py
+++ b/examples/adaptive_library/IPMRoundRotorLamMagnets.py
@@ -1,0 +1,310 @@
+# Copyright (C) 2022 - 2025 ANSYS, Inc. and/or its affiliates.
+# SPDX-License-Identifier: MIT
+#
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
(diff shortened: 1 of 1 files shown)


examples/adaptive_library/IPMRoundRotorLamMagnets.py
```diff
@@ -0,0 +1,311 @@
...
+L1_1Magnet1 = mc.get_region("L1_1Magnet1")
+L2_1Magnet2 = mc.get_region("L2_1Magnet2")
+L2_1Magnet1 = mc.get_region("L2_1Magnet1")
+
+# %%
+# Create the Adaptive Templates geometry
+# --------------------------------------
+# Use the ``round_magnet_and_pockets`` function to round the magnet regions and their corresponding
+# rotor pocket regions.
+round_magnet_and_pockets(L1_1Magnet2, rotor, L1_magnet_corner_radius, L1_rotor_lam_corner_radius)
```
JackB-Ansys, line 281: @jgsdavies It seems that any calls to Motor-CAD after this function will fail and raise the error when run with "PYMOTORCAD_DOCS_BUILD" environment variable.

examples/adaptive_library/IPMRoundRotorLamMagnets.py
```diff
@@ -0,0 +1,311 @@
...
+# If this script is loaded into the Adaptive Templates file in Motor-CAD, the current Motor-CAD
+# instance is used.
+#
+# If the script is run externally, these actions occur: a new Motor-CAD instance is opened, the e10
+# IPM motor template is loaded, the default **Corner Rounding (Rotor Lamination)** and
+# **Corner Rounding (Magnets)** is set to **None (default)** and the file is saved to a temporary
+# folder. To keep a new Motor-CAD instance open after executing the script, use the
+# ``MotorCAD(keep_instance_open=True)`` option when opening the new instance. Alternatively, use the
+# ``MotorCAD()`` method, which closes the Motor-CAD instance after the script is executed.
+# os.environ["PYMOTORCAD_DOCS_BUILD"] = "true"
```
JackB-Ansys, line 93: If you run the PY file, it runs successfully. However if you uncomment this line (or try to build the docs) it raises an error related to losing the Motor-CAD connection.

JackB-Ansys:
I think this example is ready to be merged now

examples/adaptive_library/IPMRoundRotorLamMagnets.py
```diff
@@ -0,0 +1,310 @@
...
+# Perform required imports
+# ------------------------
+# Import ``pymotorcad`` to access Motor-CAD.
+# Import ``deepcopy`` to define the adaptive template geometry.
+# Import ``os``, ``shutil``, ``sys``, and ``tempfile``
+# to open and save a temporary .mot file if none is open.
+
+from copy import deepcopy
+
+# sphinx_gallery_thumbnail_path = 'images/adaptive_templates/IPM_rounded.png'
```
james-packer: I don't think the sphinx image path should be in the middle of the imports, can it be moved above or below please?

examples/adaptive_library/IPMRoundRotorLamMagnets.py
```diff
@@ -0,0 +1,310 @@
...
+# For example: if the region to round is the rotor pocket shown below, where a line is shared with
+# a magnet region, then a list of the points circled in green will be returned. The points circled
+# in red (shared with the magnet region) are not returned by the function.
+
+# %%
+# .. image:: ../../images/adaptive_templates/IPM_rounded_1.png
+#     :width: 600pt
+
+
+def corners_to_round(region_to_round, other_regions):
```
james-packer: Could you add a comment about what this does. I think it's finding the points in region_to_round that aren't in other_regions, but it would be useful to explain this.

PR #466, opened 2025-02-04
Fix CWD when launching Motor-CAD

jgsdavies:
Fixes cwd of Motor-CAD being wrong when launching from automation. 

To replicate bug

1. Create a file with an external lab model file with a relative path:
image

2. Launch Motor-CAD using PyMotorCAD and load the file.
3. Reload model import - previously file cannot be found
image

### DIFF
--- a/src/ansys/motorcad/core/rpc_client_core.py
+++ b/src/ansys/motorcad/core/rpc_client_core.py
@@ -22,6 +22,7 @@
 
 """Contains the JSON-RPC client for connecting to an instance of Motor-CAD."""
 from os import environ, path
+from pathlib import Path
 import re
 import socket
 import subprocess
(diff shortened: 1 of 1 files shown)


src/ansys/motorcad/core/rpc_client_core.py
```diff
@@ -413,7 +414,8 @@ def _open_motor_cad_local(self):
             )
 
         motor_process = subprocess.Popen(
-            [self.__MotorExe, "/PORT=" + str(self._port), "/SCRIPTING"]
+            [self.__MotorExe, "/PORT=" + str(self._port), "/SCRIPTING"],
+            cwd=Path(self.__MotorExe).parent.absolute(),
```
jgsdavies, line 418: This sets current working directory to the parent of the Motor-CAD exe. Same as if you launched it by clicking the icon

PR #485, opened 2025-03-07
Example for stator notches as used in optimisation webinar

james-packer:
This is the template used for the NVH optimisation webinar in 2024.


examples/adaptive_library/BPMTriangularStatorNotches.py
```diff
@@ -0,0 +1,261 @@
...
+# instance will be used.
+#
+# If the script is run externally: a new Motor-CAD instance will be opened, the e10 IPM motor
+# template will be loaded and the file will be saved to a temporary folder.
+# To keep a new Motor-CAD instance open after executing the script, the option
+# ``MotorCAD(keep_instance_open=True)`` is used when opening the new instance.
+# Alternatively, use ``MotorCAD()`` and the Motor-CAD instance will close after the
+# script is executed.
+
+if pymotorcad.is_running_in_internal_scripting():
```
jgsdavies, line 94: @james-packer @JackB-Ansys  We're starting to have a lot of duplicated code in these examples. Maybe worth sticking some functions in a common unit?

also ```set_default_parameter```

jgsdavies: Not needed for this PR

examples/adaptive_library/BPMTriangularStatorNotches.py
```diff
@@ -0,0 +1,261 @@
+# Copyright (C) 2022 - 2025 ANSYS, Inc. and/or its affiliates.
```
jgsdavies, line 1: image
No thumbnail

jgsdavies (changes requested):

examples/adaptive_library/BPMTriangularStatorNotches.py
james-packer: Updated to use PyMotorCAD's built in set_adaptive_parameter_default

examples/adaptive_library/BPMTriangularStatorNotches.py
james-packer: Added sphinx_gallery_thumbnail_path, which fixes this:
image

james-packer:
I have updated this to set the region type for the notches (RegionType.stator_air). This sets up the geometry and solves correctly, but the stator notches don't currently show in the general geometry view. This is a known issue where new regions may not show up in this view if they are of a region type that Motor-CAD isn't expecting to be included from the standard geometry.

PR #487, opened 2025-03-11
Add RefreshWindingPattern call to the create_winding_pattern function.

jmacleod-ansys:
internal MotorCAD PR: https://github.com/ansys-internal/motor-cad/pull/1732

### DIFF
--- a/src/ansys/motorcad/core/methods/rpc_methods_calculations.py
+++ b/src/ansys/motorcad/core/methods/rpc_methods_calculations.py
@@ -128,5 +128,11 @@ class _RpcMethodsCalculations:
 
         Refreshes the UI to recreate winding pattern. Will be replaced by direct API call soon.
         """
+        if self.connection.check_version_at_least("2025.2"):
+            # Update parameters on the Winding Pattern page.
+            method = "RefreshWindingPattern"
+            return self.connection.send_and_receive(method)
+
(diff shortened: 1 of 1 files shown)


jmacleod-ansys:
For Jack Davies only.

src/ansys/motorcad/core/methods/rpc_methods_calculations.py
```diff
@@ -128,5 +128,11 @@ def create_winding_pattern(self):
 
         Refreshes the UI to recreate winding pattern. Will be replaced by direct API call soon.
         """
+        if self.connection.check_version_at_least("2025.2"):
```
james-packer, line 131: I think this should be 2026.1, as I think the Motor-CAD change hasn't gone into the 25R2 release?

src/ansys/motorcad/core/methods/rpc_methods_calculations.py
```diff
@@ -126,7 +126,15 @@ def do_mechanical_calculation(self):
     def create_winding_pattern(self):
         """Create winding pattern.
 
-        Refreshes the UI to recreate winding pattern. Will be replaced by direct API call soon.
+        With Motor-CAD version 27R1 or greater, updates winding pattern, properties, and weights,
```
jgsdavies: Moving these changes into a new PR. We can fix this PR later

PR #502, opened 2025-04-25
Update to round_corner and round_corners Region methods

JackB-Ansys:
Previously, you had to specify a valid **radius** for corner rounding, or an exception was raised.

Old behaviour: 

- If the **radius** is too large for the corner (due to the length of the adjacent entities) we raise an Exception ""Corner radius is too large for these entities. You must specify a smaller radius."

New behaviour:

- If the **radius** is too large for the corner (due to the length of the adjacent entities) we try to find the maximum possible **radius** that is suitable for the corner and the adjacent entities. If a new **radius** cannot be found in 100 iterations, raise the original error.

This update means that the Adaptive Templates corner rounding function behaviour is more in line with the in-built Motor-CAD Standard Template geometry corner rounding.

Note: Previously contained consolidate_lines method, now moved to separate PR #596

### DIFF
--- a/src/ansys/motorcad/core/geometry.py
+++ b/src/ansys/motorcad/core/geometry.py
@@ -102,6 +102,14 @@ class RegionType(Enum):
 RegionType.slot_area_stator_deprecated.__doc__ = "Only for use with Motor-CAD 2025.1 and earlier"
 
 
+class MagnetisationDirection(Enum):
+    """Provides an enumeration for storing Motor-CAD magnetisation direction."""
+
+    parallel = 0
+    radial = 1
+    function = 2
+
+
 class Region(object):
     """Create geometry region.
 
@@ -144,9 +152,8 @@ class Region(object):
         self._child_names = []
         self._motorcad_instance = motorcad_instance
         self._region_type = region_type
-        self.mesh_length = 0
-        self._linked_regions = []
-
+        self._mesh_length = 0
+        self._linked_region_names = []
         self._singular = False
         self._lamination_type = ""
 
@@ -401,6 +408,16 @@ class Region(object):
             new_region._br_multiplier = json["magnet_magfactor"]
             new_region._magnet_polarity = json["magnet_polarity"]
             new_region._br_magnet = json["magnet_br_value"]
+
+            if "magnetisation_direction" in json:
+                new_region.magnetisation_direction = MagnetisationDirection(
+                    json["magnetisation_direction"]
+                )
+                new_region.magnetisation_function_amplitude = json[
+                    "magnetisation_function_amplitude"
+                ]
+                new_region.magnetisation_function_angle = json["magnetisation_function_angle"]
+
         else:
             if has_region_type:
                 new_region = cls(
@@ -438,6 +455,9 @@ class Region(object):
 
         if "lamination_type" in json:
             new_region._lamination_type = json["lamination_type"]
+
+        if "linked_regions" in json:
+            new_region._linked_region_names = json["linked_regions"]
 
         new_region._raw_region = json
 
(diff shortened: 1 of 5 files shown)


src/ansys/motorcad/core/geometry.py
```diff
@@ -836,11 +871,11 @@ def round_corner(self, corner_coordinate, radius):
...
-        # lengths of the adjacent entities.
-        for index in range(len(adj_entities)):
-            j = adj_entities[index]
-            if j.length < distance:
+        # check that the  distances by which the adjacent entities are shortened are less than half
+        # the lengths of the adjacent entities.
+        for index in range(len(adj_entity_lengths)):
+            j = adj_entity_lengths[index]
+            if j < 2 * distance:
                 raise Exception(
```
james-packer, line 879: As above, maybe use ValueError instead of Exception?

src/ansys/motorcad/core/geometry.py
```diff
@@ -863,7 +898,71 @@ def round_corner(self, corner_coordinate, radius):
...
+            # round the corner
+            self._round_corner(corner_coordinate, radius, adj_entity_lengths)
+        # if the corner radius is too large, maximise the corner radius that is short enough for
+        # the adjacent entities. The adjacent entities can only be shortened to half the original
+        # entity length before any corner rounding
+        else:
+            try:
+                # try to round the corner with the specified radius
+                self._round_corner(corner_coordinate, radius, adj_entity_lengths)
+            except Exception as e:
```
james-packer, line 938: If we use ValueError instead of Exception in _round_corner() then we can instead use
`except ValueError as e:` instead of `except Exception as e:`
This would mean we would not inadvertently try to handle some unrelated exception that could be raised.

src/ansys/motorcad/core/geometry.py
```diff
@@ -863,7 +898,71 @@ def round_corner(self, corner_coordinate, radius):
...
+                    for iteration in range(100):
+                        # subtract 0.1 mm from the previous corner radius that was tried
+                        new_corner_radius -= 0.1
+                        try:
+                            # try to round the corner with the new shorter radius
+                            self._round_corner(
+                                corner_coordinate, new_corner_radius, adj_entity_lengths
+                            )
+                            break
+                        except Exception as e:
```
james-packer, line 951: Again this could be ValueError instead of Exception

src/ansys/motorcad/core/geometry.py
```diff
@@ -836,11 +871,11 @@ def round_corner(self, corner_coordinate, radius):
         if converged == False:
             raise Exception("Cannot find intersection. Check if radius is too large")
 
-        # check that the  distances by which the adjacent entities are shortened are less than the
-        # lengths of the adjacent entities.
-        for index in range(len(adj_entities)):
-            j = adj_entities[index]
-            if j.length < distance:
+        # check that the  distances by which the adjacent entities are shortened are less than half
```
james-packer, line 874: Is this being over conservative? I can think of some cases where you'd expect the original line to be shortened to less than half its original length:
image

tests/test_geometry.py
```diff
@@ -1916,6 +1994,67 @@ def test_do_not_round_corner():
...
+    assert isinstance(triangle_2.entities[3], Arc)
+    assert isinstance(triangle_2.entities[4], Line)
+    assert isinstance(triangle_2.entities[5], Line)
+    assert isinstance(triangle_2.entities[6], Arc)
+
+    # check that the point between the two parallel lines is unchanged.
+    assert triangle_2.entities[4].end == triangle_1.entities[2].midpoint
+    assert triangle_2.entities[5].start == triangle_1.entities[2].midpoint
+
+
```
james-packer: I think there are some extra ' ' (blank space) characters on line 2056 or 2057 that is upsetting the code style check!

PR #518, opened 2025-06-05
Fixed incorrect RegionType enums

aidanc-ansys:
Updated several enum types in the RegionType Class, which previously did not match with their corresponding strings in Motor-CAD and resulted in errors when users attempted to get said regions.


aidanc-ansys:
@jgsdavies is reviewing

jgsdavies (approved):

PR #519, opened 2025-06-05
Add more information to save_fea_data documentation, and move deprecation warnings so they render correctly.

james-packer:
The save_fea_data documentation wasn't very informative, so I have updated this to include a description of the parameters. Additionally, the depreciation warnings for some functions were incorrectly appearing as part of the parameter list, so these have been moved to render correctly.


james-packer:
Example of a deprecation warning before:
**
image
**
And after moving the deprecation block above the parameter list:
image

PR #525, opened 2025-06-10
Update tests to use Motor-CAD with no UI

jgsdavies:
Change Motor-CAD tests to use no UI. Should speed things up considerably

### DIFF
--- a/examples/adaptive_library/BPMTriangularRotorNotches.py
+++ b/examples/adaptive_library/BPMTriangularRotorNotches.py
@@ -102,7 +102,8 @@ else:
     mc = pymotorcad.MotorCAD(keep_instance_open=True)
     # Disable popup messages
     mc.set_variable("MessageDisplayState", 2)
-    mc.set_visible(True)
+    if not "PYMOTORCAD_DOCS_BUILD" in os.environ:
+        mc.set_visible(True)
     mc.load_template("e9")
 
     # Open relevant file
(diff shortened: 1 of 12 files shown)


examples/adaptive_library/BPMTriangularRotorNotches.py
```diff
@@ -102,7 +102,8 @@
     mc = pymotorcad.MotorCAD(keep_instance_open=True)
     # Disable popup messages
     mc.set_variable("MessageDisplayState", 2)
-    mc.set_visible(True)
+    if not "PYMOTORCAD_DOCS_BUILD" in os.environ:
```
jgsdavies, line 105: speeds things up considerably

tests/test_UI.py
```diff
@@ -22,64 +22,84 @@
...
+from RPC_Test_Common import get_temp_files_dir_path, reset_to_default_file
+from ansys.motorcad.core import MotorCAD
 
-def test_show_message(mc):
+
+@pytest.fixture(scope="module")
+def mc_ui():
+    """Set up test environment for whole unit of tests"""
+
+    os.environ["MOTORCAD_FORCE_SHOW_GUI"] = "1"
```
jgsdavies, line 35: force UI only for this module

JackB-Ansys (approved):

JackB-Ansys:
The Docs build has failed. It seems that emag_basics.py fails because it is trying to initialise_tab_names. I guess this only works with the UI?

JackB-Ansys (approved):

PR #528, opened 2025-06-11
New Region functions to construct and determine the properties of Regions

aidanc-ansys:
Implements and fixes functions to check the geometric properties of Regions, as well as a Region constructor to generate a Region when given points.

### DIFF
--- a/src/ansys/motorcad/core/geometry.py
+++ b/src/ansys/motorcad/core/geometry.py
@@ -164,6 +164,65 @@ class Region(object):
         """
         pass
 
+    @classmethod
+    def from_points_linear(
+        cls, points, region_type=RegionType.adaptive, motorcad_instance=None, sort=False
+    ):
+        """Create a region from a list of points, connecting them with lines.
+
+        Parameters
+        ----------
+        points : list of Coordinates or tuples
+        region_type : RegionType
+            Type of region
+        motorcad_instance : ansys.motorcad.core.MotorCAD
+            Motor-CAD instance currently connected
+        sort : bool
+            Reorders points to make geometry valid:
+        """
+        # conversion to Coordinates:
+        if all(isinstance(point, tuple) for point in points):
+            points_new = [0] * len(points)
+            for i, p in enumerate(points):
+                points_new[i] = Coordinate(p[0], p[1])
+            points = points_new
+
+        if sort:
+            xcenter = sum(list(point.x for point in points)) / len(points)
+            ycenter = sum(list(point.y for point in points)) / len(points)
+
+            def to_relative_coords(point):
+                return Coordinate(point.x - xcenter, point.y - ycenter)
+
+            def to_real_coords(point):
+                return Coordinate(point.x + xcenter, point.y + ycenter)
+
+            relative_points = list(map(to_relative_coords, points))
+
+            # To make sure behaviour is well-defined for points with an equal angular coordinate
+            # the points are first by sorted by radius, greatest to least
+            sorted_points_relative = sorted(
+                relative_points, key=lambda x: x.get_polar_coords_deg()[0], reverse=True
+            )
+
+            # Sort points by radial coordinate (relative to overall average position),
+            # from least to greatest angle (angles range from -180 to 180 degrees)
+            sorted_points_relative.sort(key=lambda x: x.get_polar_coords_deg()[1])
+
+            points = list(map(to_real_coords, sorted_points_relative))
+
+        reg = cls(region_type=region_type, motorcad_instance=motorcad_instance)
+
+        # Adds the lines
+        for count, point in enumerate(points):
+            if count == len(points) - 1:
+                reg.add_entity(Line(point, points[0]))
+            else:
+                reg.add_entity(Line(point, points[count + 1]))
+        if not reg.has_valid_geometry():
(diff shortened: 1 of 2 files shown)


src/ansys/motorcad/core/geometry.py
```diff
@@ -164,6 +164,65 @@ def from_coordinate_list(cls):
         """
         pass
 
+    @classmethod
+    def from_points_linear(
```
matthewAnsys, line 168: Maybe rename to include polygon? I Quite like Region.polygon()

What was your thinking for this name?

src/ansys/motorcad/core/geometry.py
```diff
@@ -164,6 +164,65 @@ def from_coordinate_list(cls):
...
+            points_new = [0] * len(points)
+            for i, p in enumerate(points):
+                points_new[i] = Coordinate(p[0], p[1])
+            points = points_new
+
+        if sort:
+            xcenter = sum(list(point.x for point in points)) / len(points)
+            ycenter = sum(list(point.y for point in points)) / len(points)
+
+            def to_relative_coords(point):
```
matthewAnsys, line 194: I would make to_relative_coords and to_real_coords part of the Coordinate class, I expect these functions will be useful outside of here.

src/ansys/motorcad/core/geometry.py
matthewAnsys: I also feel most of this should belong to the EntityList class, This is where the function name Polygon would work nicely

src/ansys/motorcad/core/geometry.py
```diff
@@ -333,19 +392,87 @@ def is_closed(self):
...
         if len(self._entities) > 0:
-            entity_first = self._entities[0]
-            entity_last = self._entities[-1]
+            # Make sure all adjacent entities have a point in common
+            return all(
+                get_entities_have_common_coordinate(self._entities[i - 1], self._entities[i])
+                for i in range(0, len(self._entities))
+            )
+        else:
+            return False
```
matthewAnsys, line 401: Can you make sure this line is tested. Should be a super easy test to validate

src/ansys/motorcad/core/geometry.py
```diff
@@ -333,19 +392,87 @@ def is_closed(self):
...
 
-            is_closed = get_entities_have_common_coordinate(entity_first, entity_last)
+    def is_nonintersecting(self):
```
matthewAnsys, line 403: I prefer is_self_intersecting() or just self_intersecting(). Also make this a property e.g.
image

src/ansys/motorcad/core/geometry.py
```diff
@@ -333,19 +392,87 @@ def is_closed(self):
...
-                    self._entities[i], self._entities[i + 1]
-                )
+        Returns
+        -------
+        Boolean
+            Whether region is nonintersecting
+        """
+        side_pairs = []
+        # Create a set of all non-neighbor pairs of lines/arcs in the region
+        for entity in self.entities:
```
matthewAnsys, line 413: Nice implementation, a line sweep algorithm would be more efficient but more complicated to implement

src/ansys/motorcad/core/geometry.py
```diff
@@ -333,19 +392,87 @@ def is_closed(self):
...
-            for i in range(len(self._entities) - 1):
-                is_closed = get_entities_have_common_coordinate(
-                    self._entities[i], self._entities[i + 1]
-                )
+        Returns
+        -------
+        Boolean
+            Whether region is nonintersecting
+        """
+        side_pairs = []
```
matthewAnsys, line 411: I prefer entity pairs over side_pairs
```suggestion
        entity_pairs = []
```

src/ansys/motorcad/core/geometry.py
matthewAnsys: I think the implementation of this should be in EntityList

src/ansys/motorcad/core/geometry.py
```diff
@@ -333,19 +392,87 @@ def is_closed(self):
...
+                    if side1.coordinate_on_entity(intersection) and side2.coordinate_on_entity(
+                        intersection
+                    ):
+                        return False
+        return True
 
-            return is_closed
-        else:
-            return False
+    def is_anticlockwise(self):
```
matthewAnsys, line 434: This is an interesting implementation, would you be able to provide info on the implementation?

tests/test_geometry.py
```diff
@@ -420,9 +445,96 @@ def test_region_to_json():
...
 
+    test_region1 = Region(region_type=RegionType.stator)
+    test_region1.add_entity(Line(Coordinate(0, 0), Coordinate(1, 1)))
+    test_region1.add_entity(Line(Coordinate(1, 1), Coordinate(1, 0)))
+    test_region1.add_entity(Line(Coordinate(1, 0), Coordinate(0.5, 0)))
+    assert test_region1.is_closed() == False
+
+
+def test_region_is_nonintersecting():
+    test_region1 = Region(region_type=RegionType.stator)
```
matthewAnsys, line 458: Can you add comments to what each of the test cases are testing e..g self intersecting or not

PR #537, opened 2025-06-19
Fix Incorrect EntityList equality evaluation

aidanc-ansys:
EntityList equality evaluation was previously evaluating whether the first list was a subset of the second. The new behaviour is as expected.

### DIFF
--- a/src/ansys/motorcad/core/geometry.py
+++ b/src/ansys/motorcad/core/geometry.py
@@ -2083,6 +2083,10 @@ class EntityList(list):
         """Compare equality of 2 EntityList objects."""
         return self._entities_same(other, check_reverse=True)
 
+    def __ne__(self, other):
+        """Compare difference of 2 EntityList objects."""
+        return not self == other
+
     def reverse(self):
         """Reverse EntityList, including entity start end coordinates."""
         super().reverse()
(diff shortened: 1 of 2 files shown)


src/ansys/motorcad/core/geometry.py
```diff
@@ -2139,6 +2139,9 @@ def _entities_same_with_direction(entities_1, entities_2):
             else:
                 return False
 
+        if len(self) != len(entities_to_compare):
+            return False
```
james-packer, line 2143: This code coverage warning is flagging up that there's a subtle problem here. Because our EntityList inherits from list, which defines an empty `__ne__(self, other)`, unless our EntityList re-defines `__ne__`, in this case `!=` is *not* the opposite of `==`, and checking `ent1 != ent2` isn't actually calling our comparison code at all!

To resolve this, I think we just need to add a new method in EntityList that is just
```
def __ne__(self, other):
    """Compare difference of 2 EntityList objects."""
    return not self == other
```

james-packer (changes requested):

src/ansys/motorcad/core/geometry.py
aidanc-ansys: Ah, thanks! I believe with that, it's now working as it should

PR #540, opened 2025-06-20
Make this more robust for different Motor-CAD versions

james-packer:
Depending on Motor-CAD versions, the region name may be the 7th or 8th item in the RegionsTable list, but always at the end. Pick the end element to make this more robust.


examples/advanced/stress_postprocessing.py
```diff
@@ -109,8 +109,8 @@ def find_divergence_point(non_linear_strain, non_linear_stress, youngs_modulus):
             non_linear_stress[i] / non_linear_strain[i], youngs_modulus, rel_tol=0.0001
         ):
             return non_linear_stress[i - 1]
-    # Return something in case we don't find an answer
-    return 0
+    # Return maximum stress in list if no divergence found
```
james-packer, line 112: Just realised that this would be a more sensible result in this edge case.

PR #541, opened 2025-06-25
Added script to create a tappered IM rotor bar

ravi-ansys:
IM rotor bar with tapered shape


ravi-ansys:
@james-packer  Hi James, I think this is now ready to merge  as passes all the checks as new example script.

PR #542, opened 2025-06-26
Add get_datastore_method and classes.

jgsdavies:
Needs: https://github.com/ansys-internal/motor-cad/pull/2557


src/ansys/motorcad/core/datastore.py
```diff
@@ -0,0 +1,216 @@
...
+            datastore_record.dynamic = json["dynamic"]
+            if datastore_record.dynamic:
+                datastore_record.array_length_2d = json["array_length_2d"]
+                datastore_record.array_length_ref_2d_name = json["array_length_ref_2d"]
+
+        else:
+            datastore_record = cls(parent_datastore)
+
+        datastore_record.current_value = json["current_value"]
+        datastore_record.default_value = json["default_value"]
```
jmacleod-ansys, line 92: I don't know if its worth having some enum or constants on the PyMotorCAD side to make these more interpretable/useful.

tests/test_rpc_client_core.py
```diff
@@ -33,6 +32,7 @@
 import ansys.motorcad.core as pymotorcad
 from ansys.motorcad.core import MotorCAD, MotorCADError, MotorCADWarning
 from ansys.motorcad.core.rpc_client_core import MOTORCAD_EXE_GLOBAL, _MotorCADConnection
+import ansys.platform.instancemanagement as pypim
```
jgsdavies, line 35: I have no idea why isort has suddenly decided to move this

src/ansys/motorcad/core/datastore.py
```diff
@@ -0,0 +1,363 @@
...
+            # Ensure the actual activex name is being removed, not the alternative.
+            self.__activex_names__.pop(record.activex_name.lower())
+            return super().pop(record.activex_name)
+        else:
+            # Keep behaviour of returning a KeyError if not in the dict.
+            raise KeyError
+
+    def extend(self, datastore: "Datastore"):
+        for key, value in datastore.items():
+            self[key] = value
```
jmacleod-ansys, line 320: I'm not completely certainly whether deep copies are needed in these cases. I think the major risk was the parent datastore, being a reference to an object, and that is hopefully mitigated by updating the parent. I have not tested if that changes the parent in the datastore you are extending from. 
This is probably also relevant for from_dict as well (not the parent bit, but pass-by-reference vs value arguments).
When I tested using deepcopy, it was taking a long time, so I hope it can be avoided.

PR #545, opened 2025-06-27
Implement GeometryTrees to contain and handle Regions

aidanc-ansys:
Implement GeometryTrees, a subclass of dict, which contain GeometryNode, a subclass of Region. Each node contains direct pointers to its parent and child nodes, avoiding additional API calls.

### DIFF
--- a/src/ansys/motorcad/core/methods/geometry_tree.py
+++ b/src/ansys/motorcad/core/methods/geometry_tree.py
@@ -182,6 +182,22 @@ class GeometryTree(dict):
         dive(node)
         return subtree
 
+    def get_nodes_from_type(self, node_type):
+        """Return all nodes in the tree of the supplied region type.
+
+        Parameters
+        ----------
+        node_type: str or RegionType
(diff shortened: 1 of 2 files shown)


src/ansys/motorcad/core/methods/adaptive_geometry.py
```diff
@@ -304,3 +305,15 @@ def reset_adaptive_geometry(self):
         # No need to do this if running internally
         if not is_running_in_internal_scripting():
             return self.connection.send_and_receive(method)
+
+    def get_geometry_tree(self):
+        """Do placeholder."""
```
JackB-Ansys, line 310: Docstrings need to be added here still I think

JackB-Ansys (approved):
I have tested this out and it seems to work nicely. I've implemented the geometry tree functionality in this Adaptive Templates example: https://github.com/ansys/pymotorcad/pull/555

JackB-Ansys (approved):

PR #558, opened 2025-07-30
fix for blackbox license parameter

HusainAdam:


src/ansys/motorcad/core/motorcad_methods.py
```diff
@@ -80,8 +80,9 @@ class MotorCAD(_MotorCADCore):
         Whether to keep the Motor-CAD instance open after the instance becomes free.
     url: string, default = ""
         Full url for Motor-CAD connection. Assumes we are connecting to existing instance.
-    use_blackbox_licence: Boolean, default: False
-            Ask Motor-CAD to consume blackbox licence.
+    use_blackbox_licence: Boolean, default: None
+        Ask Motor-CAD to consume blackbox licence. If set to None, existing Motor-CAD
+        behaviour will be used.
```
HusainAdam, line 85: Not as elegant as I like, is this acceptable? Has a value of None by default, but is also a boolean

JackB-Ansys (approved):
I think this is all good, I've tested it out with some scripts. Approving.

PR #567, opened 2025-08-08
Update pip install instructions for editable mode.

james-packer:
The old instructions to use `python -m pip install --editable ansys-motorcad-core` don't work, as the --editable option requires a path, instead of a package name. As long as we are in the correct directory, this path can just be `.`.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -131,7 +131,7 @@ To install in developer mode, complete these steps:
 
     .. code:: bash
     
-        python -m pip install --editable .
+        python -m pip install --editable .[tests,doc]
     
 #. Verify your development installation with this command:
 
--- a/README.rst
+++ b/README.rst
@@ -101,6 +101,7 @@ To install in developer mode, complete these steps:
     .. code:: bash
 
         git clone https://github.com/ansys/pymotorcad
+        cd pymotorcad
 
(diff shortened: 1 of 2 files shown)


README.rst
```diff
@@ -130,7 +131,7 @@ To install in developer mode, complete these steps:
 
     .. code:: bash
     
-        python -m pip install --editable ansys-motorcad-core
+        python -m pip install --editable .
```
jgsdavies, line 134:
```suggestion
        python -m pip install --editable .[tests,doc]
```

doc/source/contributing.rst
```diff
@@ -20,7 +20,7 @@ Run this code to clone and install the latest version of PyMotorCAD in developme
 
     git clone https://github.com/ansys/pymotorcad
     cd pymotorcad
-    pip install --editable ansys-motorcad-core
+    pip install --editable .
```
jgsdavies, line 23:
```suggestion
    pip install --editable .[tests,doc]
```

jgsdavies (changes requested):
Suggestion to add optional dependencies in instructions

doc/source/contributing.rst
james-packer: Agreed, done.

PR #587, opened 2025-09-15
Add Region and Entity methods for extending Entities (Lines and Arcs)

JackB-Ansys:
New methods for extending an entity


JackB-Ansys:
Need to add tests for extending Arcs

james-packer (approved):

PR #592, opened 2025-09-26
Add methods to save and load NVH FRFs for Motor-CAD 26R1

james-packer:


james-packer:
Marked as draft as the tests will require updated Motor-CAD version on testing machine.

PR #638, opened 2025-11-20
Best Practices for Modifying the Airgap shape

JackB-Ansys:
New section for the Adaptive Templates User Guide. I wanted to make clear that users should keep the Airgap a continuous band of equal thickness, and any modifications should be done by adding Air regions to the Rotor/Stator.

I also wanted to describe how it is necessary to add a band of air to the Rotor/Stator if we want to add narrow regions such as those shown below:

user_guide_airgap_3

The above causes issues when meshing. Instead, the user should add an air band as shown below:

user_guide_airgap_5


doc/source/user_guide/adaptive_templates.rst
```diff
@@ -675,6 +629,141 @@ The result is a modified Region with rounded corners.
...
+boundary of the group of rotor regions. The minimum thickness of the air region is 0.1 mm.
+
+.. figure:: ../images/adaptive_templates/user_guide_airgap_4.png
+    :width: 600pt
+
+    A SPM motor topology with **Magnet Reduction = 4**. A **RotorAir** region sits on top of the
+    magnet surface, forming the rotor outer boundary. The air region is 0.1 mm thick at
+    its narrowest point.
+
+For the rounded rotor notches example, an air band of specified thickness (at least 0.1 mm) should
```
james-packer: It would be useful to mention that the user will need to manually change the 'airgap' template dimension, as the template value will need to be (in this example) 0.1 mm less than the actual airgap size, so the total 'airgap' plus 'rotor air' region matches the correct value.

james-packer (changes requested):
Hi, looks really good. Just one small comment - it would be worth explicitly mentioning that the user will need to reduce the Motor-CAD template 'airgap' size to make room for the new rotor air region.

PR #641, opened 2025-11-26
Ensure we set region type in adaptive template examples

james-packer:
Also set region type to one of the ones that will show up in the radial view.

### DIFF
--- a/examples/adaptive_library/BPMTriangularRotorNotches.py
+++ b/examples/adaptive_library/BPMTriangularRotorNotches.py
@@ -237,7 +237,7 @@ for notch_loop in range(0, number_notches):
         notch_angular_width,
         notch_centre_angle,
         notch_depth,
-        region_type=RegionType.rotor_air,
+        region_type=RegionType.rotor_pocket,
     )
 
     # notch properties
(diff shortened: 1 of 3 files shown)


examples/adaptive_library/BPMTriangularRotorNotches.py
```diff
@@ -237,7 +237,7 @@
         notch_angular_width,
         notch_centre_angle,
         notch_depth,
-        region_type=RegionType.rotor_air,
+        region_type=RegionType.rotor_pocket,
```
james-packer: rotor_pocket will show up in the radial view with the e9 template, but rotor_air won't (as there are no standard regions of this type in the e9 template).

examples/adaptive_library/BPMTriangularStatorNotches.py
```diff
@@ -170,7 +170,7 @@
 
 notch_name = "Stator_Notch"
 
-notch = Region(region_type=RegionType.stator_air)
+notch = Region(region_type=RegionType.stator)
```
james-packer: I don't think stator_air is drawn (even though there are stator_air regions in the standard template), so I have changed this to stator, so we see the notches in the main radial view.

examples/adaptive_library/UpdateMaterialMesh.py
```diff
@@ -136,7 +144,7 @@
 # Create rotor band region with material and mesh properties
 # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-rt_band = Region()
+rt_band = Region(region_type=RegionType.rotor)
```
james-packer: I think this is the sensible region type for this example.

PR #643, opened 2025-11-26
Remove docs build from Python 3.9 tests.

james-packer:
The nightly build test runs on Python 3.9 to 3.14. However, the sphinx version we use isn't available for 3.9. This PR removes the docs build from the Python 3.9 tests, but still runs the main regression tests. This should avoid the nightly check routinely failing.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -9,7 +9,7 @@ isolated_build_env = build
 [gh-actions]
 description = The tox environment to be executed in gh-actions for a given python version
 python =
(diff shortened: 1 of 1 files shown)


james-packer:
To test this, I have manually triggered the nightly check on this branch, the Python 3.9 tests now pass, and run the main tests, but not the documentation build (see https://github.com/ansys/pymotorcad/actions/runs/19703157053/job/56444115231)

PR #684, opened 2026-02-24
Add option to specify model dir and accepted file dir for regression tests

ShreyaGursal:
For internal testing purposes.


jgsdavies (changes requested):

jgsdavies (changes requested):

jgsdavies (approved):

src/ansys/motorcad/core/methods/rpc_methods_testing.py
```diff
@@ -30,33 +30,61 @@ class _RpcMethodsTesting:
     def __init__(self, mc_connection):
         self.connection = mc_connection
 
-    def run_regression_tests(self, test_categories, calculation_method, decimal_separator):
+    def run_regression_tests(
+        self,
+        test_categories: str,
```
jgsdavies: Just noticed the order of these has swapped - this might cause issues for backwards compatibility

src/ansys/motorcad/core/methods/rpc_methods_testing.py
```diff
@@ -30,33 +30,61 @@ class _RpcMethodsTesting:
...
         -------
         dict
             Test results including name, test case, diff type, and major and minor differences.
         """
         method = "RunRegressionTests"
-        params = [test_categories, calculation_method, decimal_separator]
+
+        # Remove trailing None values
+
+        params = params = [
```
jgsdavies: For 26r1 branch I guess this will pull PyMotorCAD main and then break RuNRegression.py? Might be worth adding a version check in here so only versions >= 2027 runs these extra args?

I guess this is what you were trying to do previously by using the extra code to pop the extra args. Sorry, missed this connection!

jgsdavies (changes requested):

PR #705, opened 2026-03-18
Update load_dxf_file for 2027R1

jgsdavies:
Goes with Motor-CAD PR 6329


matthewAnsys (dismissed):

James-Schaffer:
Coverage can be ignored for now, feature does not exist on version of MotorCAD on the runners. Local testing results :

Running tests on main (feature doesn't exist so tests legacy functionality) :
image

Running tests on this branch :
image

PR #718, opened 2026-03-30
Add oil cooling calibration example

JackB-Ansys:


HusainAdam:
HI @JackB-Ansys  I have added a first draft of this example. See below for a full list of things to do:

TODO:
* ~~Fix Image sizes~~ done
* ~~Fix location of test_cases.csv and general folder structure (working folder ----> inputs folder and outputs folder)~~ done
* ~~Update test_cases.csv data to create better results~~ done
* Update tolerance value and default number of workers
* ~~Modify implementation to allow to solve during docs build (?) Currently all of script is encapsulated to not run. Not sure how to do this cleanly without duplicating functionality, and if we _should_ be running this example during docs build due to considerable execution time.~~ Done, calibration procedure longer runs during docs build
* Include additional postprocessing showing results before vs after and calibration factors

I wonder if we should have this as a static example that does nothing during the docs run. If we want this to run each time, it will be very slow especially if we want to use a useful number of test cases. This would also make it easier to avoid having to create a special case to get the docs build to work.

examples/advanced/oil_cooling_calibration.py
```diff
@@ -0,0 +1,1070 @@
...
+# ejected onto the front and rear end windings in equal proportions. There are 12 nozzles at the
+# front and another 12 nozzles at the rear, with each nozzle being 1.5mm in diameter. In our initial
+# Motor-CAD model, a Housing Water Jacket is also present, though this is not used in the
+# experimental testing.
+#
+# .. image:: ../../images/oil_cooling_calibration/spray_fluid_flow.png
+#    :align: center
+#
+# The Motor-CAD Spray Cooling Radial (from Housing) cooling system makes use of heat transfer
+# correlations to determine the heat transfer coefficient (HTC) of the oil cooling. The correlations
```
JackB-Ansys: Maybe we should use 'h' instead of 'HTC' for consistency with the Motor-CAD UI

PR #726, opened 2026-04-16
Example - oil cooling calibration implementation

JackB-Ansys:
Updates to oil cooling calibration example.

Add implementation, where new Correlation Factors are set in Motor-CAD based on cooling parameters - implements calibrated model.


JackB-Ansys:
To do:

- Explanation of enamel temperatures (image from webinar series)
- Improve calibration messages
- Add validation for one of the test cases, compare temperatures - bar chart or graph showing each hairpin layer temp
- Implementation using HWJ, different speed and flow rates from test cases

HusainAdam:
TODO: 

- Update tolerance value and default number of workers
- Add another csv file link
- Update image of csv file
- Verify blackbox license functionality and implement if appropriate
- Update csv file to have non-calibration closer to calibrated results

JackB-Ansys:
To do:
- BlackBox licence clarification
- Create separate enamel temperature function within PyMotorCAD
- Internal Motor-CAD script setup

PR #731, opened 2026-04-22
Add api to offset region (requires updated Motor-CAD API)

james-packer:
It would be useful to be able to offset regions (inflate/deflate) for example to make margins around an existing region. This PR enables access to the Motor-CAD API for this.


JackB-Ansys:
This seems to be struggling in the case where we have rounded corners. Using the e4a template IPM rotor pocket regions, I get good results if I turn off the standard template corner rounding:
image
image
However if I enable the standard template corner rounding, it seems to be incorrectly creating some of the arc entities: 
image
It looks like in some of these cases the arcs are being created with the inverse radius (radius * -1)

jgsdavies (approved):

PR #732, opened 2026-04-28
Update tests duct geometry method

matthewAnsys:
## Description

Updates the geometry tests to handle both the original and improved duct geometry methods (`DuctGeometryMethod`).

### Changes

- **`test_region_children`**: Updated the assertion on the number of rotor children to be conditional on the active `DuctGeometryMethod` variable — the original method (`0`) produces 16 children, while the improved method (`1`) produces 15.

- **`test_region_linked_regions`**: Wrapped the test in a `try/finally` block that temporarily sets `DuctGeometryMethod` back to `0` (original) before asserting linked regions, then resets it to `1` (improved) in the `finally` block. `reset_adaptive_geometry()` is called after each variable change to ensure the geometry is regenerated correctly.


James-Schaffer:
Skip in 26R2
image

Pass in latest with checkiffeatureexists
image

PR #743, opened 2026-05-27
Add context to geometry_export API

matthewAnsys:
### Description

Updates the `geometry_export` method to support an optional `context` parameter, allowing users to specify which analysis context to export the geometry for.

### Changes

- **`rpc_methods_general.py`**: Updated `geometry_export` to accept an optional `context` string parameter (`"Magnetic"`, `"Thermal"`, or `"Mechanical"`). The method now calls the new `GeometryExportWithContext` RPC method instead of `GeometryExport`, passing the context as a parameter. Defaults to an empty string (current context behaviour).

- **`test_general.py`**: Updated the geometry export test to call `geometry_export("Magnetic")` on Motor-CAD 2027.0+, falling back to the no-argument call on earlier versions.

### Motivation

The previous `geometry_export()` API had no way to specify which context's geometry to export. This change exposes the `GeometryExportWithContext` RPC introduced in Motor-CAD 27R1, enabling context-aware geometry export.


tests/test_general.py
```diff
@@ -156,7 +156,11 @@ def test_export_multi_force_data(mc):
 def test_geometry_export(mc):
     file_path = get_temp_files_dir_path() + r"\dxf_export_file.dxf"
     mc.set_variable("DXFFileName", file_path)
-    mc.geometry_export()
+
```
jmacleod-ansys: A bit of whitespace that needs to be trimmed/Black run. Same the docstring of geometry_export.

jmacleod-ansys (changes requested):

jmacleod-ansys (dismissed):

PR #744, opened 2026-06-04
Add  `context` parameter to `check_if_geometry_is_valid` API

matthewAnsys:
## Summary

Adds an optional `context` parameter to `check_if_geometry_is_valid` to support Motor-CAD 2027R1+, which introduces a new RPC endpoint (`CheckIfGeometryIsValidWithContext`) requiring an explicit geometry context.

## Changes

- **`rpc_methods_geometry.py`**: Updated `check_if_geometry_is_valid` to accept a `context` parameter (`"Magnetic"`, `"Thermal"`, or `"Mechanical"`). For Motor-CAD 2027.0+, the context is required and the new `CheckIfGeometryIsValidWithContext` RPC method is called. For earlier versions, behavior is unchanged.
- **`rpc_client_core.py`**: Extended the special-case success handling to also recognise `CheckIfGeometryIsValidWithContext` alongside `CheckIfGeometryIsValid`.
- **test_geometry.py**: Updated `test_check_if_geometry_is_valid` to call the correct API variant based on the connected Motor-CAD version.

## Notes

- A `MotorCADError` is raised if `context` is omitted when connected to Motor-CAD 2027R1 or later, as the legacy endpoint is no longer available in that version.

---

### DIFF
--- a/src/ansys/motorcad/core/methods/rpc_methods_geometry.py
+++ b/src/ansys/motorcad/core/methods/rpc_methods_geometry.py
@@ -21,7 +21,7 @@
 # SOFTWARE.
 
 """RPC methods for geometry."""
-from ansys.motorcad.core.rpc_client_core import MotorCADError
+from ansys.motorcad.core.rpc_client_core import MotorCADWarning
 
 
 class _RpcMethodsGeometry:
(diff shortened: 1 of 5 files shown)


src/ansys/motorcad/core/methods/rpc_methods_geometry.py
```diff
@@ -99,12 +100,27 @@ def check_if_geometry_is_valid(self, edit_geometry):
...
             ``1`` if an attempt to reset the geometry has been made, ``O`` otherwise.
         """
-        method = "CheckIfGeometryIsValid"
-        params = [edit_geometry]
+        if self.connection.check_version_at_least("2027.0"):
+            if context == "":
+                raise MotorCADWarning(
+                    "It is recommended to specify the context for check_if_geometry_is_valid"
+                    " with Motor-CAD 2027.0 or later. If no context is specified, the geometry"
+                    " will be checked for the current UI context, this not work for headless mode."
```
jgsdavies, line 118:
```suggestion
                    " will be checked for the current UI context, this will not work for headless mode."
```

PR #765, opened 2026-07-02
Add requests session as beta feature

jgsdavies:


jgsdavies:
Merging this to test Motor-CAD fix - only enabled by beta env var anyway

PR #779, opened 2026-07-10
Added rpc command for getting used node to node resistance

James-Schaffer:
Added get_node_to_node_resistance_used()

Description :

 - get_node_to_node_resistance() returns the resistance between 2 nodes without taking into account the resistance multiplier.
 - new method get_node_to_node_resistance_used() takes into account resistance multiplier.


src/ansys/motorcad/core/methods/rpc_methods_thermal.py
```diff
@@ -188,6 +188,25 @@ def get_node_to_node_resistance(self, node1, node2):
         params = [node1, node2]
         return self.connection.send_and_receive(method, params)
 
+    def get_node_to_node_resistance_used(self, node1, node2):
```
jgsdavies: Not sure this is a good name, maybe Matt R has some better ideas. Also can we add this to the original function instead of creating a new one? My preference would be a parameter that defaults to the original result. Maybe we could have:

- a new parameter with a nice name. Maybe ```include_resistance_multiplier```?
- get_node_to_node_resistance calling your new API method alongside the old one.
- Give the user the new/old result depending on ```include_resistance_multiplier```
- Warn the user if they don't match and they haven't explicity set the new parameter to True or False

Be sure to also consider the case where a user calls this method on an old version of Motor-CAD without the API method. We don't want to break this!

jgsdavies (changes requested):

James-Schaffer:
I need to update testing script.

src/ansys/motorcad/core/methods/rpc_methods_thermal.py
```diff
@@ -178,14 +181,53 @@ def get_node_to_node_resistance(self, node1, node2):
...
+                warnings.warn(
+                    "Resistance multiplier is being used but "
+                    "include_resistance_multiplier is not set. "
+                    "Defaulting to old method.",
+                    stacklevel=2,
+                )
+            return resistance
+
+        # If in date and set, do method requested
+        elif include_resistance_multiplier is True:
```
jgsdavies, line 224: I think the logic is correct but a little hard to read. Any ways to make this nicer?

jgsdavies (changes requested):

tests/test_thermal.py
```diff
@@ -56,6 +56,73 @@ def test_get_node_to_node_resistance(mc):
     assert almost_equal(res, 0.0043, 3)
 
 
+def test_get_node_to_node_resistance_old(mc, monkeypatch):
+    monkeypatch.setattr(mc.connection, "program_version", "2027.1.0.1")
```
jgsdavies: Do we need to monkeypatch this every time? Locally we should have a version that is later than 27.0. If we do run this on an older version we will just hit an exception in Motor-CAD.
What if we replaced this with:

```suggestion
    if not mc.connection.check_version_at_least("2027.0"):
        pytest.skip("Requires Motor-CAD 2027.0")
```
Then the tests will pass locally and the runner will skip them until Motor-CAD is updated.

jgsdavies (changes requested):

tests/test_thermal.py
James-Schaffer: This sounds sensible to me. I'm going to make these changes for 2 tests which need the latest version.

PR #781, opened 2026-07-16
Add split region around entity api

James-Schaffer:
Added split_region_about_entity()
Added _to_json() for Line and Arc entities

Description :
 - split_region_about_entity() takes in a region and an entity, splits the region into pieces about the entity passed in and then returns them in a list
 - region.split_about_entity() does the same but updates self to a split region and only returns the other split regions


James-Schaffer:
Coverage can be ignored, tests skipped due to feature not being on current runner motorcad build.

Skips on current release (feature doesn't exist) :
image


Passes on newest build (feature exists) : 
image

James-Schaffer:
Skips on current release (feature doesn't exist) :
image


Passes on newest build (feature exists) :
image

matthewAnsys (approved):

PR #797, opened 2026-08-06
Update the subtract region testing to work with unique region naming

James-Schaffer:
Subtract region now enforces regions returned from it have unique names.


James-Schaffer:
Current build
image

26R2
image

matthewAnsys (approved):

PR #798, opened 2026-08-06
Add delay to quit

dougiehawkins:
Add 0.5 sec delay to quit of local instance.
 **test_keeping_instance_open** was failing as Motor-CAD instance was not fully cleaned up before a new connection to the instance was attempted. This delay has been added to ensure the instance has fully closed.


src/ansys/motorcad/core/rpc_client_core.py
```diff
@@ -780,4 +816,8 @@ def _quit(self):
         else:
             # local machine
             method = "Quit"
-            return self.send_and_receive(method)
+            result = self.send_and_receive(method)
```
jmacleod-ansys: Please revert everything except just adding the sleep. The rest will get added in another PR.

jmacleod-ansys (changes requested):

tests/test_rpc_client_core.py
jmacleod-ansys, line 182: I would just add to this comment to say quit should be called and no MotorCAD open.

tests/test_rpc_client_core.py
```diff
@@ -180,12 +180,14 @@ def test_keeping_instance_open(monkeypatch):
...
     # Check keep_instance_open ignored when building docs
+    # quit will now be called and Motor-CAD will close if instance is deleted
     monkeypatch.setenv("PYMOTORCAD_DOCS_BUILD", "True")
 
     mc2 = MotorCAD(keep_instance_open=True)
 
     original_port = mc2.connection._port
-
+	
+    # as PYMOTORCAD_DOCS_BUILD has been set then this should call quit and Motor-CAD should close
```
jmacleod-ansys:
```suggestion

```

jmacleod-ansys (dismissed):

tests/test_rpc_client_core.py
dougiehawkins: done

jmacleod-ansys (approved):

jgsdavies (approved):

PR #801, opened 2026-08-12
Adaptive Templates Excample - IPM Rotor Pocket adjustment

JackB-Ansys:
A general script that identifies the vertices of rotor pocket ends, and creates adaptive parameters for adjusting the shape based on the vertices as control points.

Control points can be shifted with a distance coordinate and a direction coordinate.

Individual corner radii can be applied to each control point vertex.

An option to include the midpoints of rotor pocket end entities as additional control points. This gives greater degrees of freedom.

Together, the point shifting and corner rounding allows a wide range of custom rotor pocket end shapes can be generated, either manually or by running in conjunction with an optimisation workflow.


JackB-Ansys:
To do:

- Add section that checks whether modified pockets reduce the bridge or post thickness. 
- This could be optional, enabled with a boolean/if statement
- Example of creating 'utility regions' such as a 'bridge band' or 'post strip'
- Subtract the utility regions from the modified pockets to preserve the bridge and post
- Work out how to handle rounded corners when the bridge/post region is subtracted

PR #802, opened 2026-08-12
Updated test_get_region_dxf to check load_dxf_file_with_context

James-Schaffer:
Test uses mc.load_dxf_file which needs to pass a context in the newest release. This checks if the feature exists and if so passes the context.


James-Schaffer:
Passes on current motorcad build
image

Passes on current release
image

jgsdavies (approved):

PR #805, opened 2026-08-14
Wait for MotorCAD process to quit correctly on Linux

jmacleod-ansys:
Add code to improve handling of quit and make sure the MotorCAD process has actually quit before exiting. This is more of an issue on Linux where it can become a zombie if wait is not called.


ravi-ansys (approved):

james-packer (approved):
Reviewed code, looks sensible.

PR #807, opened 2026-08-18
Multi region subtraction

James-Schaffer:
Adding the ability to subtract multiple regions from a region in one go.

Using the same method Region.subtract() but passing an list of Regions will trigger it to subtract all of them from the target region.


James-Schaffer:
Need to update tests to ensure it works with unique region naming

James-Schaffer:
Tested and passed on ensure_unique_region_names_for_api_methods branch, 26R1 and current main.

image

James-Schaffer:
Passing all tests on main, nightlybuild and 26R1 still.

PR #810, opened 2026-08-19
Deprecate get_licence functions

Kaiser-E:
The functions get_licence and get_license both do not do anything as the Motor-CAD licence is checked out during startup. Deprecate the functions so that old scripts still work.

### DIFF
--- a/src/ansys/motorcad/core/methods/rpc_methods_general.py
+++ b/src/ansys/motorcad/core/methods/rpc_methods_general.py
@@ -494,7 +494,7 @@ class _RpcMethodsGeneral:
           This method is deprecated. License is checked out at startup.
         """
         if self.connection.check_version_at_least("2027.1"):
-            print('get_licence is deprecated for 27R1 or later. License is checked out at startup.')
+            print("get_licence is deprecated for 27R1 or later. License is checked out at startup.")
         else:
             method = "GetLicence"
             return self.connection.send_and_receive(method)
--- a/src/ansys/motorcad/core/methods/rpc_methods_general.py
+++ b/src/ansys/motorcad/core/methods/rpc_methods_general.py
@@ -490,9 +490,15 @@ class _RpcMethodsGeneral:
         """Check if a license is available for the current context and machine type.
 
         If such a license is available, it is checked out.
+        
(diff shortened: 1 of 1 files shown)


src/ansys/motorcad/core/methods/rpc_methods_general.py
```diff
@@ -490,9 +490,15 @@ def get_licence(self):
         """Check if a license is available for the current context and machine type.
 
         If such a license is available, it is checked out.
+
```
jmacleod-ansys:
```suggestion

```

src/ansys/motorcad/core/methods/rpc_methods_general.py
```diff
@@ -490,9 +490,15 @@ def get_licence(self):
...
 
         If such a license is available, it is checked out.
+        
+        ..note::
+          This method is deprecated. License is checked out at startup.
         """
-        method = "GetLicence"
-        return self.connection.send_and_receive(method)
+        if self.connection.check_version_at_least("2027.1"):
+            print('get_licence is deprecated for 27R1 or later. License is checked out at startup.')
```
jmacleod-ansys:
```suggestion
            print("get_licence is deprecated for 27R1 or later. License is checked out at startup.")
```

jmacleod-ansys (changes requested):
Formatter changes (make sure pre-commit is installed and implemented)

src/ansys/motorcad/core/methods/rpc_methods_general.py
```diff
@@ -490,9 +490,15 @@ def get_licence(self):
         """Check if a license is available for the current context and machine type.
 
         If such a license is available, it is checked out.
+
+        ..note::
+          This method is deprecated. License is checked out at startup.
         """
-        method = "GetLicence"
-        return self.connection.send_and_receive(method)
+        if self.connection.check_version_at_least("2027.1"):
```
jmacleod-ansys, line 497: This will break a bunch of licensing tests. These will need to be updated. check_if_feature_exists should help.

jmacleod-ansys (changes requested):

PR #815, opened 2026-08-25
Fix for pytests hanging and remove skip licencing tests

James-Schaffer:
Ensure motorcad is left in a state where popups/messages are turned off after test_message_config and test_verbose_message_config. Also update conftest.py to use new motorcad message config commands when the feature exists.

These seem to be have causing the issues with the tests.

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -77,9 +77,13 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip tox tox-gh-actions
-          
+
       - name: Test with tox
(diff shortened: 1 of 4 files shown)


tests/test_rpc_client_core.py
```diff
@@ -406,6 +406,7 @@ def test_show_gui(mc):
         MotorCAD(open_new_instance=False, port=mc.connection._port, show_gui=True)
 
 
+# this is causing issues , mem access violation
```
James-Schaffer:
```suggestion
# this sometimes was causing issues , mem access violation, when running locally on win64
```

jgsdavies (approved):

PR #816, opened 2026-08-26
Check for environment variable indicating running in internal scripting

jmacleod-ansys:
Add a fix for a failure to detect internal scripting is running for some adaptive template files. This is not super easy for Linux, aside from setting something in the environment and detecting that.


james-packer (approved):
Looks sensible, and shouldn't affect other functionality.

PR #811, opened 2026-08-19
Document which PyMotorCAD commands are available for Linux 

Caroline-Leak:
Adds a page to the User Guide with a list of which commands are available for Linux for 27R1, which are not available for 27R1 and which have not yet been tested.

PR #822, opened 2026-09-02
Rename restore_compatibility_settings

Caroline-Leak:
Rename restore_compatibility_settings to restore_deprecated_settings to match changes to naming of settings in Motor-CAD. 

Note: there is an added reference to restore_compatibility_settings in linux_scripting.rst in #811, which will need updating in whichever of the two PRs merges second.


src/ansys/motorcad/core/methods/rpc_methods_variables.py
```diff
@@ -69,7 +69,22 @@ def set_array_variable_2d(self, array_name, array_index1, array_index2, new_valu
...
+        """
+        warn(
+            "restore_compatibility_settings is deprecated. Use restore_deprecated_settings",
+            DeprecationWarning,
+        )
+        method = "RestoreCompatibilitySettings"
+        return self.connection.send_and_receive(method)
+
+    def restore_deprecated_settings(self, deprecated=True, recommended=False):
+        """Restore model deprecated settings to deprecated values to use the old methods."""
```
jmacleod-ansys: Repetition of docstring. This should be removed.

jmacleod-ansys (changes requested):

src/ansys/motorcad/core/methods/rpc_methods_variables.py
Caroline-Leak: good spot, this also doesn't make sense as written.

src/ansys/motorcad/core/methods/rpc_methods_variables.py
```diff
@@ -69,7 +69,22 @@ def set_array_variable_2d(self, array_name, array_index1, array_index2, new_valu
...
         """Restore model compatibility settings to default values to use the latest methods.
```
Caroline-Leak:
```suggestion
```

src/ansys/motorcad/core/methods/rpc_methods_variables.py
Caroline-Leak:
```suggestion
        """Restore model deprecated settings to default values to use the old methods."""
```

src/ansys/motorcad/core/methods/rpc_methods_variables.py
```diff
@@ -69,8 +69,22 @@ def set_array_variable_2d(self, array_name, array_index1, array_index2, new_valu
...
+        """
+        warn(
+            "restore_compatibility_settings is deprecated. Use restore_deprecated_settings",
+            DeprecationWarning,
+        )
+        method = "RestoreCompatibilitySettings"
+        return self.connection.send_and_receive(method)
+
+    def restore_deprecated_settings(self, deprecated=True, recommended=False):
+        """Restore model deprecated settings to default values to use the old methods."""
```
Caroline-Leak: Made another mistake in the last update of this
```suggestion
        """Restore model deprecated settings to default values to use the latest method."""
```

src/ansys/motorcad/core/methods/rpc_methods_variables.py
jmacleod-ansys: There's a """ that should be removed as it's broken the file.

jmacleod-ansys (changes requested):

jmacleod-ansys (approved):

PR #826, opened 2026-09-07
Wait for MotorCAD process to quit correctly on Linux

jmacleod-ansys:
Add code to improve handling of quit and make sure the MotorCAD process has actually quit before exiting. This is more of an issue on Linux where it can become a zombie if wait is not called.

Windows continues to have problems with quitting, so this is only done for Linux at this time. It is hoped to perform this for Windows as well in time, but that does not seem to yet be a thing.


src/ansys/motorcad/core/rpc_client_core.py
```diff
@@ -895,11 +898,59 @@ def get_last_error_message(self):
...
+        """
         if self.pim_instance is not None:
             self.pim_instance.delete()
         else:
             # local machine
+            if not psutil.pid_exists(self.pid):
+                # The process has already exited, so send_and_recieve will fail.
+                # Possible that another MotorCAD object has already sent the Quit command,
+                # or the user has closed Motor-CAD.
+                raise MotorCADError(
```
james-packer: Do we want to raise an exception here? In `__del__` above, we intentionally don't pass on exceptions.

src/ansys/motorcad/core/rpc_client_core.py
```diff
@@ -895,11 +898,59 @@ def get_last_error_message(self):
...
-    def _quit(self):
-        """Quit MotorCAD."""
+    def _quit(self, max_wait=200):
+        """Quit MotorCAD.
+
+        Parameters
+        ----------
+        max_wait : int, optional
+            Maximum number of seconds to wait for the Motor-CAD process to exit before force
+            killing it. Default is 200.
```
james-packer: Maybe worth mentioning that the 'force quit' is only for linux (currently)?

src/ansys/motorcad/core/methods/rpc_methods_general.py
```diff
@@ -527,9 +527,16 @@ def clear_message_log(self):
...
-        """Quit Motor-CAD."""
-        self.connection._quit()
+    def quit(self, max_wait=200):
+        """Quit Motor-CAD.
+
+        Parameters
+        ----------
+        max_wait : int, optional
+            Maximum number of seconds to wait for the Motor-CAD process to exit before force
+            killing it. Default is 200.
```
james-packer:
```suggestion
            killing it (Note: This argument only has an effect on Linux). Default is 200.
```

james-packer (approved):

PR #829, opened 2026-09-09
Update doc generation

James-Schaffer:
`autofill_function_names.py` is old and not built to support the new syntax of mc.[unit].<method>. With the addition of `mc.messageconfig.<method>` methods, the doc builder needed to be updated to support this and be future proofed a bit.

Changes :
 - Updated `autofill_function_names.py` script :
   - Now uses an array of _DocCategory dataclass to represent the categories instead of 2 lists
   - Added _DocCategoryGenType which can be expanded, this allows different categories to generate docs in different ways.
   - Added Unit categories, a new type which supports the `mc.messageconfig.<method>` synax
- Re-structured how templates work, now can be found in `.../doc/source/_templates` folder, contains a mix of static and generated templates used to generate the .rst files.
  - Added templates for default, unit categories and for unit methods.
- Updated comments to make additions in future easier

### DIFF
--- a/doc/source/methods/autofill_function_names.py
+++ b/doc/source/methods/autofill_function_names.py
@@ -12,6 +12,15 @@ import shutil
 #     _DocCategory obj.
 #  2. Run this script to generate the documentation files.
 #
+# If you are adding a Unit category spesificly, the rpc_methods_core python file will
+# need to be updated. A class attribute will need to be added and set to be an alias of
+# the rpc class, this attribute will share the name as the runtime instance used. This is
+# used by the doc gen to find the methods but has no functional impact when using the api.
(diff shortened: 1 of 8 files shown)


doc/source/methods/autofill_function_names.py
```diff
@@ -1,100 +1,201 @@
...
-        "rpc_methods_ui.py",
-        "rpc_methods_utility.py",
-        "rpc_methods_variables.py",
+    # Docs can be generated in different ways depending on the category type :
+    # default   :   MotorCAD.<method>
+    # unit      :   MotorCAD.[unit].<method>
+    #
+    # See _DocCategoryGenType for structure
+    #
+    doc_catagories = [
```
JackB-Ansys: typo - should be 'doc_categories'

PR #830, opened 2026-09-11
Add warnings for gui/Windows only functions

jgsdavies:
Add warnings for gui/Windows only functions

for @Caroline-Leak @jmacleod-ansys

### DIFF
--- a/src/ansys/motorcad/core/rpc_client_core.py
+++ b/src/ansys/motorcad/core/rpc_client_core.py
@@ -798,6 +798,33 @@ class _MotorCADConnection:
             else:
                 success_value = _METHOD_SUCCESS
 
+            # Post 2027R1 - When supported is False, the server skipped
+            # the call because it is not available on the current Motor-CAD
+            # platform (headless or Linux).
+            if "supported" in response["result"] and response["result"]["supported"] is False:
+                # functionScope enum: 0=ftGuiOnly, 1=ftWindowsOnly, 2=ftAllPlatforms, 3=ftUndefined
+                scope = response["result"].get("functionScope")
+                if response["result"]["supported"] == 0:
+                    scope_available = "This function is only available in Motor-CAD with a GUI."
+                elif response["result"]["supported"] == 1:
+                    scope_available = "This function is only available in Motor-CAD on Windows."
+                else:
+                    # Server should only set supported=False for the scopes above.
+                    self._raise_if_allowed(
+                        f"'{method}' returned supported=False with unexpected "
+                        f"functionScope={scope!r}."
+                    )
+                    return
+                warnings.warn(
+                    f"'{method}' was skipped. {scope_available}",
+                    MotorCADWarning,
+                )
+                # Server forces success=kSuccess when skipping the call.
+                if success != success_value:
+                    self._raise_if_allowed(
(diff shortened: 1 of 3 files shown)


src/ansys/motorcad/core/rpc_client_core.py
```diff
@@ -801,6 +798,33 @@ def send_and_receive(self, method, params=None, success_var=None):
...
+                elif response["result"]["supported"] == 1:
+                    scope_available = "This function is only available in Motor-CAD on Windows."
+                else:
+                    # Server should only set supported=False for the scopes above.
+                    self._raise_if_allowed(
+                        f"'{method}' returned supported=False with unexpected "
+                        f"functionScope={scope!r}."
+                    )
+                    return
+                warnings.warn(
```
Caroline-Leak, line 818: Do we also need to set scope_available in the else case above? Will the warning still be shown in that case?

Caroline-Leak (changes requested):

src/ansys/motorcad/core/rpc_client_core.py
```diff
@@ -801,6 +801,34 @@ def send_and_receive(self, method, params=None, success_var=None):
...
+                    f"'{method}' was skipped. {scope_available}",
+                    MotorCADWarning,
```
jmacleod-ansys, line 824: Do we want to return for any method that was skipped? I doubt we can trust any output if it is unsupported.

jmacleod-ansys (changes requested):

src/ansys/motorcad/core/rpc_client_core.py
jgsdavies: We should still return some Nones in the correct format e.g. ```None, None``` for tuples or an empty list

jmacleod-ansys (dismissed):

jmacleod-ansys (approved):

tests/test_rpc_client_core.py
```diff
@@ -315,6 +315,13 @@ def test_warnings(mc, monkeypatch):
         mc.get_variable("n/a")
 
 
+def test_unsupported_method_warning(mc_headless):
```
jmacleod-ansys, line 318: Can we up the coverage by having a test that a fully supported method doesn't give a warning?

jmacleod-ansys (changes requested):

tests/test_rpc_client_core.py
```diff
@@ -315,6 +315,22 @@ def test_warnings(mc, monkeypatch):
...
+    # In fully headless Motor-CAD the server marks GUI-only methods as unsupported
+    # and skips them. The client should emit a MotorCADWarning without raising.
+    with pytest.warns(MotorCADWarning, match="only available in Motor-CAD with a GUI"):
+        mc_headless.set_visible(False)
+
+
+def test_supported_method_no_warning(mc_headless):
+    with warnings.catch_warnings(record=True) as caught_warnings:
+        warnings.simplefilter("always")
+        result = mc_headless.get_variable("MessageDisplayState")
```
jmacleod-ansys: This will give a deprecation warning. Can we get a different parameter?

jmacleod-ansys (approved):

PR #832, opened 2026-09-16
Update examples for motor modified templates

james-packer:
Required to work with some minor geometry changes. Also tidy up some title underlining, to reduce how many warnings are generated.


samples/adaptive_template/TaperedBar.py
```diff
@@ -57,7 +54,11 @@ def chord_angle(cord_length, r):
 # Set IM motor type if not already
 if not pymotorcad.is_running_in_internal_scripting():
     mc.load_template("i6a")
+    # Simplify template geometry by turning off bar corner rounding
+    mc.set_variable("Bar Corner Radius[T]", 0)
 
+# Reset geometry to default
```
james-packer, line 60: Just moved this so it's after the set_variable command above, so the geometry in Motor-CAD gets updated.

examples/adaptive_library/DXFImport.py
```diff
@@ -111,6 +111,13 @@
     mc.set_array_variable(
         "RotorCircularDuctLayer_ChannelDiameter", 1, 4.94
     )  # Set the Rotor Duct Diameter to 4.94 mm for Layer 2
+    mc.set_variable("CornerRounding_Rotor", 0)  #  Turn off corner rounding
```
james-packer, line 114: The template now has corner rounding enabled, so needs to be turned off to get the same regions as before.

examples/advanced/force_extraction.py
```diff
@@ -76,8 +76,9 @@
 #
 #   - All motor types and options: `LoadPoint_Speed_Array`
 #   - BPM or similar for torque based definition: `LoadPoint_Torque_Array`
-#   - BPM or similar for current/phase advance definition: `LoadPoint_Current_Array`,
-#   and `LoadPoint_PhaseAdvance_Array`
+#   - BPM or similar for current/phase advance definition:
```
james-packer, line 79: The auto formatter had split this line, which broke the flow of the table, I've changed it so these are on nested bullet points, instead of one long line.

samples/adaptive_template/TaperedBar.py
```diff
@@ -57,7 +54,11 @@ def chord_angle(cord_length, r):
 # Set IM motor type if not already
 if not pymotorcad.is_running_in_internal_scripting():
     mc.load_template("i6a")
+    # Simplify template geometry by turning off bar corner rounding
+    mc.set_variable("Bar Corner Radius[T]", 0)
```
james-packer, line 58: Again, the template model now has corner rounding on, so turn this off to get back to the simple bar geometry.

JackB-Ansys (approved):

